PR #3, opened 2015-06-02
add desired cap validation

Jonahss:
depends upon https://github.com/appium/node-mobile-json-wire-protocol/pull/9

### DIFF
--- a/lib/desired-caps.js
+++ b/lib/desired-caps.js
@@ -1,0 +1,102 @@
+import log from './logger';
+import validator from 'validate.js';
+
+let desiredCapabilityConstraints = {
+  platformName: {
+    presence: true,
+    isString: true,
+    inclusion: [
+      'iOS',
+      'Android',
+      'FirefoxOS'
+    ]
+  },
+  deviceName: {
+    presence: true,
+    isString: true
+  },
+  platformVersion: {},
+  newCommandTimeout: {
+    isNumber: true
+  },
+  automationName: {
+    inclusion: [
+      'Appium',
+      'Selendroid'
+    ]
+  },
+  autoLaunch: {
+    isBoolean: true
+  },
+  udid: {
+    isString: true
+  },
+  orientation: {
+    inclusion: [
+      'LANDSCAPE',
+      'PORTRAIT'
+    ]
+  },
+  autoWebview: {
+    isBoolean: true
+  },
+  noReset: {
+    isBoolean: true
+  },
+  fullReset: {
+    isBoolean: true
+  }
+};
+
+validator.validators.isString = function (value) {
+  if (typeof value === 'string') {
+    return null;
+  }
+
+  if (typeof value === 'undefined') {
+    return null;
+  }
+
+  return `must be of type string`;
+};
+validator.validators.isNumber = function (value) {
+  if (typeof value === 'number') {
+    return null;
+  }
+
+  if (typeof value === 'undefined') {
+    return null;
+  }
+
+  return `must be of type number`;
+};
+validator.validators.isBoolean = function (value) {
+  if (typeof value === 'boolean') {
+    return null;
+  }
+
+  if (typeof value === 'undefined') {
+    return null;
+  }
+
+  return `must be of type boolean`;
+};
+validator.validators.isObject = function (value) {
+  if (typeof value === 'object') {
+    return null;
+  }
+
+  if (typeof value === 'undefined') {
+    return null;
+  }
+
+  return `must be of type object`;
+};
+validator.validators.deprecated = function (value, options, key) {
+  if (options) {
+    log.warn(`${key} is a deprecated capability`);
+  }
+  return null;
+};
+
+export { desiredCapabilityConstraints, validator };
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -2,18 +2,37 @@ import { MobileJsonWireProtocol,
          errors } from 'mobile-json-wire-protocol';
 import commands from './commands';
 import log from './logger';
+import { desiredCapabilityConstraints, validator } from './desired-caps';
 import B from 'bluebird';
 import _ from 'lodash';
 
+validator.promise = B;
+validator.prettify = (val) => { return val; };
+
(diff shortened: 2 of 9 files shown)


Jonahss:
Tests fail because  appium/node-mobile-json-wire-protocol#9 hasn't been merged and then published

lib/desired-caps.js
```diff
@@ -0,0 +1,102 @@
+import log from './logger';
+import validator from 'validate.js';
+
+validator.validators.isString = function(value) {
```
sebv: Looks like, there is a scoping problem here, we should fork the library and fix module export so that we can create an instance of validator detached from the npm cache.

jlipps: we do `function (value) {`

i'm also surprised that these basic type validations aren't built into the library

lib/driver.js
```diff
@@ -2,18 +2,38 @@ import { MobileJsonWireProtocol,
...
+    // base-driver internals
     this.curCommand = new B((r) => { r(); }); // see note in execute
     this.noCommandTimer = null;
-    this.newCommandTimeoutMs = null;
+    this.shouldValidateCaps = shouldValidateCaps;
+  }
+
+  // we only want subclasses to ever extend the contraints
+  set desiredCapConstraints (constraints) {
+    this._constraints = _.merge(this._constraints, constraints);
```
jlipps: why `_.merge` and not `Object.assign` (which is the new thing we're using)

lib/driver.js
```diff
@@ -23,8 +43,43 @@ class BaseDriver extends MobileJsonWireProtocol {
...
+                                 _.keys(this._constraints));
+    if (extraCaps.length) {
+      log.warn(`The following capabilities were provided, but are not ` +
+               `recognized by appium: ${extraCaps.join(', ')}.`);
+    }
+  }
+
   validateDesiredCaps (caps) {
-    return !!caps;
+    if (!this.shouldValidateCaps) { return true; }
```
jlipps:
```js
if (!this.shouldValidateCaps) {
  return true;
}
```

lib/driver.js
```diff
@@ -23,8 +43,43 @@ class BaseDriver extends MobileJsonWireProtocol {
...
+
+    let validationOptions = { fullMessages: false };
```
jlipps:
```js
let validationOptions = {fullMessages: false};
```

(spaces in between braces are only for imports/exports)

jlipps:
love the approach! and the tests :-)

i made some comments, also if we need all the validator functions can we put them in their own file?

jlipps:
btw mjsonwp has been merged so feel free to publish at your leisure

lib/desired-caps.js
sebv: Yeah but it's like:
```
require('validate.js').validators.isString = function(value)
```
So our custom functions are hanging from the npm cache. Cleaner way in node would be:
```
validator = require('validate.js')()
validator.validators.isString = function(value)
```

test/capability-specs.js
```diff
@@ -0,0 +1,219 @@
...
+      'deviceName': 'Delorean',
+      'extra': 'cheese',
+      'hold the': 'sauce'
+    });
+  });
+
+  it('should log the use of extra caps', async function () {
+    this.timeout(500);
+
+    let spy = sinon.spy(logger, 'warn');
```
sebv: Just wondering should we do the init in `before` blocks?

test/capability-specs.js
```diff
@@ -0,0 +1,219 @@
...
+
+    await d.createSession({
+      'platformName': 'iOS',
+      'deviceName': 'Delorean',
+      'extra': 'cheese',
+      'hold the': 'sauce'
+    });
+
+    spy.callCount.should.be.above(0);
```
sebv: Same for all test, you can just do:
`logger.warn.callCount.should.be.above(0);`

lib/desired-caps.js
Jonahss: gotcha. That will work because when drivers which extend BaseDriver create their own constraints objects, they don't need a reference to these functions, just use the name as a key. neat.

lib/driver.js
Jonahss: originally I wasn't cloning the default contraits, so object.assign was mutating the default for all further drivers. Now that I'm cloning, object.assign works again :)

test/capability-specs.js
Jonahss: hm ok. And put them in a separate suite?

Jonahss:
Corrections made.

Following @sebv's advice to not add the validator functions to the npm cache, so I exported the validator from the function (it does not have a constructor).

@jlipps was asking about exporting the validator functions, but the interesting mechanism here is that users just add a key in their constraints object which gets matched to the list of validators connected to the validator object. So they never need a reference to the actual functions in order to use them.

If they need a reference to them, they can just import the validator from desired-caps.js (but really never should have to).

jlipps:
I wasn't really talking about exporting the validator functions. I was talking about 2 things:

1. surprise that you had to fill out these basic type validations (that they weren't included in validator.js)
2. a request that the type validations be moved to their own file (not so that they could be exported, just so that they didn't clutter up the caps logic)

jlipps:
so yeah, apart from moving the validator functions into their own file, I'm :+1:

Jonahss:
Ah.
1. Yeah I was super surprised as well. The reason is because the module was
made for validating web forms.
2. ok, but are they good where they are? I don't think there's much clutter
since that's all that file will ever hold. Just validators and the one
constraints object which uses them.

Jonahss:
Ah. So you really do want to move them out? I felt that it was clearer to have them here, since this is the only place they are used, and users who implement other drivers would want to know where they come from.

Otherwise the file would only hold this single constraints object. That what you want?

jlipps:
i guess they're fine where they are. it's a little weird to come into the file and scroll through a list of validations before you see the actual purpose of the file. maybe we could move the validations to the bottom instead?

Jonahss:
Oh yeah, that works for me :)

jlipps:
k feel free to merge when that is done then

PR #5, opened 2015-08-07
exposing configureApp

sebv:
Not sure if it is the best way, but we need it.

ping @jlipps


jlipps:
i think it should already be available as `this.helpers.configureApp` to anything that extends `BaseDriver`?

Jonahss:
Ooh, yeah `this.helpers.configureApp`

PR #6, opened 2015-08-17
Adding new device-settings

moizjv:
- @jlipps @Jonahss please review

### DIFF
--- a/lib/device-settings.js
+++ b/lib/device-settings.js
@@ -1,0 +1,38 @@
+import _ from 'lodash';
+import log from './logger';
+
+class DeviceSettings {
+
+  constructor (defaultSettings = {}) {
+    this._settings = defaultSettings;
+  }
+
+  // calls updateSettings from implementing driver every time a setting is changed.
+  async update (newSettings) {
+    if (!_.isObject(newSettings)) {
+      throw new Error("Settings update should be called with valid JSON");
+    }
+    for (let prop of _.keys(newSettings)) {
+      if(this._settings[prop] !== newSettings[prop]) {
+        let updatePayload = {
+          key: prop,
+          value: newSettings[prop],
+          oldValue: this._settings[prop],
+        };
+        this._settings[prop] = newSettings[prop];
+        // update setting only when there is updateSettings defined.
+        if (this.onSettingsUpdate) {
+          await this.onSettingsUpdate(updatePayload);
+        } else {
+          log.errorAndThrow("Unable to update settings; onSettingsUpdate method not found");
+        }
+      }
+    }
+  }
+
+  getSettings () {
+    return this._settings;
+  }
+}
+
+export default DeviceSettings;
--- a/test/driver-e2e-tests.js
+++ b/test/driver-e2e-tests.js
@@ -6,6 +6,7 @@ import chai from 'chai';
 import chaiAsPromised from 'chai-as-promised';
 import 'mochawait';
 import B from 'bluebird';
+import DeviceSettings from '../lib/device-settings';
 
 const should = chai.should();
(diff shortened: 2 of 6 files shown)


lib/device-settings.js
```diff
@@ -0,0 +1,40 @@
+import _ from 'lodash';
+import log from './logger';
+
+class DeviceSettings {
+
+  constructor() {
+    this._settings = {};
+    // this is where default settings can be declared
+    this._settings.ignoreUnimportantViews = false;
```
jlipps: we should probably expose defaults as an argument to the constructor so that each driver can define them. basedriver shouldn't know about android views

lib/driver.js
```diff
@@ -39,6 +40,7 @@ class BaseDriver extends MobileJsonWireProtocol {
     this.shutdownUnexpectedly = false;
     this.noCommandTimer = null;
     this.shouldValidateCaps = shouldValidateCaps;
+    this.settings = new DeviceSettings();
```
jlipps: we should then pass in default settings to the driver constructor, _or_ require each driver to init its own DeviceSettings obj

lib/device-settings.js
```diff
@@ -0,0 +1,40 @@
...
+          key: prop,
+          value: newSettings[prop],
+          oldValue: this._settings[prop],
+        };
+        this._settings[prop] = newSettings[prop];
+        // update setting only when there is updateSettings defined.
+        if (this.updateSetting) {
+          await this.updateSettings(updatePayload);
+        } else {
+          log.errorAndThrow("Unable to update settings updateSetting method not found");
```
jlipps: "Unable to update settings; updateSetting method not found"

jlipps:
couple comments, otherwise LGTM. but how about the actual mjsonwp command? shouldn't it go in the commands list? (i.e., handling the user set and get settings should be in base driver too)

Jonahss:
code looks fine to me.

Jonahss:
ah I agree with jlipps. Gotta add routes to mjsonwp

moizjv:
@Jonahss added commands :)

lib/device-settings.js
```diff
@@ -0,0 +1,38 @@
+import _ from 'lodash';
+import log from './logger';
+
+class DeviceSettings {
+
+  constructor(_settings = {}) {
```
jlipps: no reason for the local variable to also have an underscore

test/driver-e2e-tests.js
```diff
@@ -176,7 +177,20 @@ function baseDriverE2ETests (DriverClass, defaultCaps = {}) {
...
     describe('settings api', () => {
-      // TODO port over settings tests
+      before(() => {
+        d.settings = new DeviceSettings({ignoreUnimportantViews: false});
+      });
+      it('should be able to get settings object',() => {
+        d.settings.getSettings().ignoreUnimportantViews.should.be.false;
+      });
+      it('should throw error when updateSettings method is not defined', async () => {
+        await d.settings.update({ignoreUnimportantViews: true}).should.eventually
```
jlipps: `should.eventually.be.rejectedWith`

might as well make it grammatical :-)

moizjv:
done

lib/device-settings.js
```diff
@@ -0,0 +1,38 @@
+import _ from 'lodash';
+import log from './logger';
+
+class DeviceSettings {
+
+  constructor(settings = {}) {
```
jlipps: `constructor (`

jlipps: also, let's call the local variable `defaultSettings` so it's more clear what this is

PR #10, opened 2015-08-18
add generic Find commands, and locator strategy validation

Jonahss:


lib/commands/timeout.js
```diff
@@ -0,0 +1,13 @@
+import log from '../logger';
+
+let commands = {}, helpers = {}, extensions = {};
+
+commands.implicitWait = function (ms) {
+  this.implicitWaitMs = parseInt(ms);
```
jlipps: `parseInt(ms, 10);`

PR #11, opened 2015-08-18
Adding onSettingsUpdate as parameter to constructor

moizjv:


Jonahss:
Ah i see. I still like events, but if you want to do callback this should be fine.

PR #13, opened 2015-08-20
command timeout

Jonahss:


sebv:
lgtm.

lib/commands/timeout.js
```diff
@@ -7,6 +7,12 @@ commands.implicitWait = function (ms) {
   log.debug(`Set implicit wait to ${ms}ms`);
 };
 
+commands.timeouts = function (name, duration) {
```
jlipps: would still like to see this decorated with `async` because that's what MJSONWP requires. in this case it's just creating an instantly-resolving promise of course.

Jonahss: ok, commiting.

PR #16, opened 2015-08-25
Extra validation

sebv:
Added them so I don't see the warning at startup. @imurchie can you help setting those the way it is intended, I am not too familiar the api yet.


imurchie:
These should go into the ios driver. I'm working on validating `launchTimeout`, since it can be a number of different things. See https://github.com/appium/appium-ios-driver/pull/5

PR #17, opened 2015-08-26
Adds reset + command timeout improvements

sebv:
I think it is enough for iOS, so much simpler. ping @jlipps @moizjv 

Android has extra logic (or use different) on top of that, I'll let Moiz figure it out.

### DIFF
--- a/lib/commands/session.js
+++ b/lib/commands/session.js
@@ -24,6 +24,10 @@ commands.createSession = async function (caps) {
     this.newCommandTimeoutMs = (this.caps.newCommandTimeout * 1000);
   }
 
+  // We need to ininitialize one onUnexpectedShutdow promise per session
+  // to avoid the promise fulfilment being propagated between sessions.
+  this.resetOnUnexpectedShutdown();
+
   log.info(`Session created with session id: ${this.sessionId}`);
 
   return [this.sessionId, caps];
@@ -47,11 +51,8 @@ commands.getSession = async function () {
 };
 
 commands.deleteSession = async function () {
-  if (this.noCommandTimer) {
-    clearTimeout(this.noCommandTimer);
-  }
   this.opts = this.initialOpts;
-  this.noCommandTimer = null;
+  this.clearNewCommandTimeout();
   this.sessionId = null;
 };
 
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -36,14 +36,27 @@ class BaseDriver extends MobileJsonWireProtocol {
     // base-driver internals
     this.curCommand = new B((r) => { r(); }); // see note in execute
     this.curCommandCancellable = new B((r) => { r(); }); // see note in execute
-    this.onUnexpectedShutdown = new B(() => {}).cancellable();
-    // handle this ourselves so we don't get a stacktrace printed out
-    this.onUnexpectedShutdown.catch(() => {});
     this.shutdownUnexpectedly = false;
     this.noCommandTimer = null;
     this.shouldValidateCaps = shouldValidateCaps;
     // settings should be instantiated by implementing drivers
     this.settings = null;
+    this.resetOnUnexpectedShutdown();
+  }
+
+  /*
+   * Initialize a new onUnexpectedShutdown promise, cancelling existing one.
+   */
+  resetOnUnexpectedShutdown() {
+    // TODO: use positive logic (i.e. onShutdown)
+    if (this.onUnexpectedShutdown && !this.onUnexpectedShutdown.isFulfilled()) {
+      this.onUnexpectedShutdown.cancel();
+    }
+    this.onUnexpectedShutdown = new B((resolve, reject) => {
+      this.unexpectedShutdownDeferred  = {resolve, reject};
+    }).cancellable();
+    // noop handler to avoid warning.
+    this.onUnexpectedShutdown.catch(() => {});
   }
 
   // we only want subclasses to ever extend the contraints
@@ -106,9 +119,7 @@ class BaseDriver extends MobileJsonWireProtocol {
   async executeCommand (cmd, ...args) {
     // if we had a command timer running, clear it now that we're starting
     // a new command and so don't want to time out
-    if (this.noCommandTimer) {
-      clearTimeout(this.noCommandTimer);
-    }
+    this.clearNewCommandTimeout();
 
     // if we don't have this command, it must not be implemented
     if (!this[cmd]) {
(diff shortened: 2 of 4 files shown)


sebv:
more fixes...

sebv:
Related to: https://github.com/appium/appium/pull/5535

sebv:
+ better shutdown logic.

lib/commands/session.js
```diff
@@ -13,6 +13,7 @@ commands.createSession = async function (caps) {
   this.validateDesiredCaps(caps);
   this.sessionId = UUID.create().hex;
   this.caps = caps;
+  this.rawCaps = _.cloneDeep(caps); // used for reset
```
jlipps: i thought we already had this somewhere?

lib/commands/session.js
```diff
@@ -24,6 +25,8 @@ commands.createSession = async function (caps) {
     this.newCommandTimeoutMs = (this.caps.newCommandTimeout * 1000);
   }
 
+  this.resetOnUnexpectedShutdown();
```
jlipps: don't understand why we are calling this in createSession

lib/driver.js
```diff
@@ -149,25 +156,31 @@ class BaseDriver extends MobileJsonWireProtocol {
     // automatic session deletion in this.onCommandTimeout. Of course we don't
     // want to trigger the timer when the user is shutting down the session
     // intentionally
+    let onCommandTimeout = async function () {
```
jlipps: i swear i've done this before:

```js
let onCommandTimeout = async () => {
```

lib/commands/session.js
sebv: because you can reject promise only once.

lib/driver.js
sebv: Yes I moved it, and fixed the promise logic, there was a sync function as a a parameter of setTimeout.

lib/commands/session.js
jlipps: I realize I started this whole concept, but it's starting to get pretty confusing. I think we need some better comments or explanation about what this and other code is doing.

lib/commands/session.js
sebv: We copy initialOpts 3 line below, but they are not caps. Maybe we should merge the 2.

Jonahss: we have `this.cap` and we never mutate it. we copy from caps to opts

lib/driver.js
```diff
@@ -184,8 +197,27 @@ class BaseDriver extends MobileJsonWireProtocol {
...
+
+    log.debug("Resetting app mid-session");
+    log.debug("Running generic full reset");
+    let currentConfig = {};
+    for (let k of ['implicitWaitMs', 'newCommandTimeoutMs', 'sessionId']) {
+      currentConfig[k] = this[k];
+    }
+    await this.deleteSession();
+    log.debug("Restarting app");
+    await this.createSession(this.rawCaps);
```
Jonahss: you can just use `this.caps` right? pretty sure we don't mutate that object.

lib/commands/session.js
sebv: Thanks @Jonahss , looks like I'll need to search/replace `this.caps` by `this.opts` in a lot of places.

lib/commands/session.js
sebv: I'll put some comments tomorrow,

lib/driver.js
sebv: yeah looks like it.

lib/commands/session.js
jlipps: ah yes. here was the vision for `this.caps` and `this.opts`:

`this.caps` was supposed to be exactly what the user sent in. `this.opts` starts out as any server arguments or options that the driver is instantiated with. Because we want caps to be able to override driver options, we clone caps and merge it onto `this.opts`. So throughout the driver, we should usually only ever use `this.opts`. `this.caps` exists merely as a reference to what the user _actually_ sent it and should never be mutated.

jlipps: we should probably add a functional test that each driver runs that asserts that after a test session, `this.caps` remains unchanged.

also, we must not forget to import the basedriver functional tests into each other driver so that they can run the same tests (basedriver functional tests should pass for each sub-driver too!)

sebv: Things would be easier to manage if we had session objects and the drivers were stateless handlers.

jlipps: hmm, good idea. so each command would get a session object that it could use.

sebv:
Made some fix added comments.

lib/driver.js
```diff
@@ -184,8 +194,29 @@ class BaseDriver extends MobileJsonWireProtocol {
...
+    let currentConfig = {};
+    for (let k of ['implicitWaitMs', 'newCommandTimeoutMs', 'sessionId']) {
+      currentConfig[k] = this[k];
+    }
+    await this.deleteSession();
+    log.debug("Restarting app");
+    await this.createSession(this.caps);
+
+    for (let [k,v] of _.pairs(currentConfig)) {
+      this[k] = v;
```
jlipps: I wonder if there's a bug here when `createSession` begins proxying to something else (like Chromedriver or Selendroid), because the JWProxy object will be told that our current sessionId is whatever new one `createSession` comes up with, rather than the one we are trying to retain in `currentConfig`. We should test this case and make sure it's not a bug. If it is, we might want to be able to pass in a sessionId to `createSession` so it uses that one instead of coming up with a new one.

jlipps:
OK @sebv this is a lot more clear. I added one comment which we can turn into a TODO if you want to get this merged.

lib/driver.js
sebv: I was talking with this with @moizjv the other day, and in my opinion it is easier and safer to generate all the session in AppiumDriver, and always use those. If a component is using another session it should store it internally, and we can probably expose it in the /session whenever the customer needs it.

jlipps: you're saying that appium should generate a sessionId and pass it to the new sub-driver session it is instantiating?

jlipps: I'm down with that idea. so we'd do what I just suggested and allow `createSession` to take a sessionId parameter so it doesn't make its own.

sebv: Yeah we can do that for now, but I don't think the device driver even need to now about their session id. There are some cases like if we need it for logging and responding to /session, but we can do that in AppiumDriver.

jlipps: appium-android-driver needs to be able to have session ids so it can stand as a driver on its own, right?

sebv: You mean like chromedriver?

In this case it should just have a simple server embedded within it's code, not within the class itself.

jlipps: It could work. We do need it in a few places as you said but maybe there's a more elegant way. Would be kind of nice to get that state out of the driver. That means we need to do session state checking in the webserver somehow (or mjsonwp)

PR #19, opened 2015-08-26
Adding get swipe options

moizjv:


sebv:
This should probbably live in a gesture.js command file.

moizjv:
@jlipps and I had a talk about it. It is more of a helper function not a command

sebv:
We should devise a way so that everything doesn't ends up in driver.js and the code is sort of functionally organized.

jlipps:
LGTM. @sebv it's either this or we make a new helpers file and add it to the prototype. either way, but so far this is the only thing like it for android driver so i suggested putting it on the driver

sebv:
It's similar to https://github.com/appium/appium-android-driver/blob/master/lib/commands/find.js#L11, we already have helpers in command file for this. Anyway not too important, we can reformat later.

jlipps:
yes, it's exactly the same as that, we just didn't want to make a new file immediately. but i agree it can be factored out easily. NBD.

moizjv:
Merging it, will refactor later if needed

PR #24, opened 2015-08-28
simpler download logic

sebv:
Had already fixed it, it is the same as the current solution, but more idiomatic. 

ping @jlipps @Jonahss

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -7,7 +7,7 @@ import B from 'bluebird';
 import { tempDir, system, fs } from 'appium-support';
 import { exec } from 'teen_process';
 import AdmZip from 'adm-zip';
-import requestWithoutPromises from 'request';
+import request from 'request';
 
 const ZIP_EXTS = ['.zip', '.ipa'];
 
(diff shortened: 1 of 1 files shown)


lib/helpers.js
```diff
@@ -56,26 +56,18 @@ async function downloadApp (app, appExt) {
...
-  let onClose = B.promisify(writeStream.on, writeStream)('close');
+  let targetPath = await tempDir.path({prefix: 'appium-app', suffix: suffix});
 
-  requestWithoutPromises.get(fileUrl).pipe(writeStream);
-
-  await onClose;
-  logger.debug(`${fileUrl} downloaded to ${fileInfo.path}`);
-  return fileInfo.path;
+  // don't use request-promise here, we need streams
+  await new B(function(resolve, reject) {
```
jlipps: arrow function?

jlipps:
i do think this is more idiomatic. @Jonahss do you agree?

Jonahss:
Ooh yes I like it better.

PR #29, opened 2015-09-01
reset fix

sebv:
Now the problem is that we create a new onUnexpectedShutdown promise during reset, and cancel the old one, resulting in the session not being removed from the list on crashes.

We need to preserve resetOnUnexpectedShutdown during reset. The current logic being a little bit hardcore, just calling createSession+deleteSession, it is simpler to just disable the function for the duration of the rest.

Long term solution would be to extract a Session object and have a special constructor for reset.

ping @jlipps.


jlipps:
:+1:

PR #30, opened 2015-09-05
Better command timeouts

sebv:
ping @jlipps 
It's more flexible this way, and there cannot be 2 timeouts running at the same time.


sebv:
Merging, or ios driver will be broken.

PR #33, opened 2015-10-07
Isaac cleanup

imurchie:


lib/commands/timeout.js
```diff
@@ -29,22 +31,22 @@ helpers.startNewCommandTimeout = function (cmd) {
...
-    B.resolve(onCommand()).done();
-  }, this.newCommandTimeoutMs);
+    })
+    .catch(B.CancellationError, (/*err*/) => {
+      // ignore
+    });
 };
 
-helpers.implicitWaitForCondition = async function (condFn) {
+helpers.implicitWaitForCondition = async (condFn) => {
```
jlipps: I'm curious how this works. We needed to not use arrow functions before since they bind the function when we want to bind it when we add these helper functions to the prototype.

lib/driver.js
```diff
@@ -178,10 +178,7 @@ class BaseDriver extends MobileJsonWireProtocol {
     return res;
   }
 
-  async startUnexpectedShutdown (err = null) {
-    if (err === null) {
-      err = new errors.NoSuchDriverError("The driver was unexpectedly shut down!");
-    }
+  async startUnexpectedShutdown (err = new errors.NoSuchDriverError('The driver was unexpectedly shut down!')) {
```
jlipps: i'm surprised this works. cool.

jlipps:
wow this was a big cleanup :+1:

PR #38, opened 2015-11-30
Allow booleans to be strings

imurchie:
Our old behaviour was to allow boolean caps to be sent as strings (e.g., `"false"` being the same as `false` as a cap value). Continue that tradition.


jlipps:
test to make coverage happy?

PR #40, opened 2016-01-20
pass sessionId arg when calling deleteSession. Make error message for…

Jonahss:
… NewCommandTimeout more helpful

### DIFF
--- a/lib/commands/timeout.js
+++ b/lib/commands/timeout.js
@@ -24,7 +24,7 @@ helpers.clearNewCommandTimeout = function () {
   }
 };
 
-helpers.startNewCommandTimeout = function (cmd) {
+helpers.startNewCommandTimeout = function () {
   // make sure there are no rogue timeouts
   this.clearNewCommandTimeout();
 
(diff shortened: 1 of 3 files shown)


imurchie:
Where is the `sessionId` being used in `deleteSession`? It is just in the `AppiumDriver`? If so, can you add the parameter to the `BaseDriver` version as a placeholder, so it is documented that it is passed in?

```
commands.deleteSession = async function (/* sessionId */) {
// ...
```

lib/commands/timeout.js
```diff
@@ -36,7 +36,10 @@ helpers.startNewCommandTimeout = function (cmd) {
     .then(async () => {
       log.warn(`Shutting down because we waited ` +
                `${this.newCommandTimeoutMs / 1000} seconds for a command`);
-      await this.startUnexpectedShutdown(new Error(`Command ${cmd} timed out!`));
+      let errorMessage = `New Command Timeout of ${this.newCommandTimeoutMs}` +
```
jlipps: if we're going to advertise, we should say that it's seconds, since in this log we show that it is in ms, and that could lead to confusion

jlipps:
one comment then LGTM

PR #41, opened 2016-01-20
update dependencies, most notably: lodash to v4

Jonahss:
Couldn't update Bluebird, it will require more than a couple minutes. Might have to be updated appium-wide.

Anyways, created a ticket for bluebird, but updated lodash and others. Thank goodness for tests.


imurchie:
Does `lodash` need to be updated throughout?

Jonahss:
@imurchie lodash needs to be updated for everybody, but we can do it package-by-package, since we never share the `_` variable between modules.

imurchie:
Ah, ok. That makes sense.

This looks good, then.

PR #43, opened 2016-01-22
Adding case insensitive inclusion validator

moizjv:


lib/desired-caps.js
```diff
@@ -110,6 +110,17 @@ validator.validators.deprecated = function (value, options, key) {
   }
   return null;
 };
+validator.validators.inclusionCaseInsensitive = function (value, options) {
+  if (typeof value !== 'string') {
+    return `unrecognised`;
```
imurchie: This can be a plain old string. Also, what output comes when this happens?

moizjv: for consistency with above used strings i added that quotes. Output will be like ` platformName unrecognised.`

imurchie: Ah. That is a good output. 

As for the strings, the rest should be changed rather than another be added in the wrong format.

imurchie:
Looks good!

Any ideas about the failing download of node?

moizjv:
@imurchie ok now it passed; though coverage is complaining, is it good to merge?

imurchie:
Haha! Coverage went up and coveralls is complaining.

Yes, merge away.

PR #49, opened 2016-05-06
pull other packages into base-driver

jlipps:
according to our dearchitecture plan.

This also involved updating readmes, writing some tests, using new lodash, etc...

Given the many changes, if you're curious, might want to look commit-by-commit


jlipps:
hmm, running the jsonwp-proxy tests on their own in travis to see if it's something that might have changed in travis.

jlipps:
ahh it's because we weren't running e2e tests for jsonwp-proxy before

jlipps:
ok @imurchie once this passes build it should be ready to merge assuming we're not intending to make any minor changes to base-driver right now, or any changes at all to the mjsonwp, jwproxy, jwstatus, or express packages! (until we go through and update the deps for all packages which used to depend on them)

imurchie:
Cool! I look forward to getting that all updated!

imurchie:
👍

PR #50, opened 2016-05-17
status is not always successful.

rafael-chavez:


rafael-chavez:
From my testing all the responses the webdriver are attached to a [status](https://github.com/facebook/WebDriverAgent/blob/master/WebDriverAgentLib/Routing/FBCommandStatus.h) 

Unless there is some error on the webdriver side the command response should have a status on it. I can add a guard to make it have a status like:
```
httpResBody.status = driverRes.status == null ? JSONWP_SUCCESS_STATUS_CODE : driverRes.status;
```

rafael-chavez:
As is this logic is incorrect as for example, a command that returns with a status of 7 (element not found) will return to the client with a status of 0, which is incorrect.

lib/mjsonwp/mjsonwp.js
```diff
@@ -245,8 +245,8 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
         driverRes = null;
       }
 
-      // assuming everything worked, set up the json response object
-      httpResBody.status = JSONWP_SUCCESS_STATUS_CODE;
+      // Response status should be the status set by the driver response.
+      httpResBody.status = driverRes.status;
```
jlipps: how about:

```js
httpResBody.status = _.isUndefined(driverRes.status) ? JSONWP_SUCCESS_STATUS_CODE :
                                                       driverRest.status;
```

to ensure that a poorly-constructed server we're proxying to doesn't cause a failure?

jlipps:
good catch!

lib/mjsonwp/mjsonwp.js
rafael-chavez: I like this. updating!

jlipps:
@rafael-chavez looks good, however (a) can you please squash and (b) it looks like your git author is not set to match your github e-mail or something? can you double-check that?

imurchie:
Looks good with the check. Android and iOS drivers both don't return a status for any command, since we rely on throwing errors in error conditions. Proxied commands are the exception, so this ought to catch those and be good! Thanks!

rafael-chavez:
@jlipps i just squashed commits. As far as the git author, i did a 

```
git config user.email myemail
```

is that what your asking?

jlipps:
@rafael-chavez is it the same as your github e-mail? github is not showing a link to your user under the commit author. or maybe you need to re-push somehow?

rafael-chavez:
yea i think that was it (i needed to re-push) i see my little picture icon now. 

screen shot 2016-05-19 at 8 57 22 am

jlipps:
@rafael-chavez looks like this broke some tests, maybe there was a different expectation in the past:

```
  1) MJSONWP via express router should handle commands with no response values:
     StatusCodeError: 500 - [object Object]
      at new StatusCodeError (node_modules/request-promise/lib/errors.js:26:15)
      at Request.RP$callback [as _callback] (node_modules/request-promise/lib/rp.js:68:32)
      at Request.self.callback (node_modules/request/request.js:200:22)
      at Request.<anonymous> (node_modules/request/request.js:1067:10)
      at IncomingMessage.<anonymous> (node_modules/request/request.js:988:12)
      at endReadableNT (_stream_readable.js:913:12)
  2) MJSONWP via drivers jsonwp proxy should avoid proxying deleteSession commands:
      AssertionError: expected 500 to equal 200
      + expected - actual
      -500
      +200
      
      at Object.callee$2$0$ (test/mjsonwp/mjsonwp-specs.js:662:29)
      at tryCatch (node_modules/babel-runtime/regenerator/runtime.js:67:40)
      at GeneratorFunctionPrototype.invoke [as _invoke] (node_modules/babel-runtime/regenerator/runtime.js:315:22)
      at GeneratorFunctionPrototype.prototype.(anonymous function) [as next] (node_modules/babel-runtime/regenerator/runtime.js:100:21)
      at GeneratorFunctionPrototype.invoke (node_modules/babel-runtime/regenerator/runtime.js:136:37)
      at bound (domain.js:287:14)
      at GeneratorFunctionPrototype.runBound (domain.js:300:12)
      at run (node_modules/babel-runtime/node_modules/core-js/library/modules/es6.promise.js:104:47)
      at node_modules/babel-runtime/node_modules/core-js/library/modules/es6.promise.js:115:28
      at flush (node_modules/babel-runtime/node_modules/core-js/library/modules/$.microtask.js:19:5)
```

you should be able to run the tests locally with `gulp once` and `gulp e2e-test` to debug

rafael-chavez:
i'll take a look @jlipps

rafael-chavez:
@jlipps there was an edge case that i needed to cover _.isNil(driverRes) and there was a test that needed to get updated in proxy-specs.js. Done.

jlipps:
LGTM now. @imurchie ?

test/jsonwp-proxy/proxy-specs.js
```diff
@@ -134,18 +134,19 @@ describe('proxy', () => {
...
         e = err;
       }
-      should.exist(e);
-      e.message.should.contain('Original error: Invisible element');
-      e.value.should.eql({message: 'Invisible element'});
-      e.status.should.equal(11);
+      should.exist(res);
+      res.value.should.eql({message: 'Invisible element'});
+      res.status.should.eql(11);
+      should.not.exist(e);
```
imurchie: This check for `e`, along with the try/catch block are not necessary, since this does not throw anymore. We just need

```
let j = mockProxy({sessionId: '123'});
let res = await j.command('/element/200/text', 'GET');
res.value.should.eql({message: 'Invisible element'});
res.status.should.eql(11);
```

rafael-chavez: let me remove.

rafael-chavez:
updated @imurchie !

imurchie:
Looks good! Thanks for all the work, and your patience with the process!

PR #51, opened 2016-05-31
Add comment for find command return type

Jonahss:
The exact specification of what to return in a findElement command was a little hard to find. This was the first place I looked, so a good place to put the answer.

@imurchie for review :)


imurchie:
Good idea!

Should there also be an analogous comment for the `findElements` method? Returns an array of elements in the same form as `findElement`, and an empty array if no elements are found?

jlipps:
👍

PR #53, opened 2016-06-20
Allow data: scheme for protractor tests

### LINKED ISSUE #52: Allow 'data:' scheme in setUrl validator
Protractor uses url `data:text/html,<html></html>` to defer Angular bootstrap, but this URL isn't valid according to the `setUrl` method.

Each test fails with error `Url or Uri must start with <scheme>://`

imurchie:
Fixes #52


jlipps:
👍

PR #54, opened 2016-06-21
Make sure the server is closed when we call close

imurchie:
We have cases (notably within the xcuitest driver) where we get odd results because the server is not closing correctly.

This is backwards compatible, in that an un-awaited promise behaves the same as a function that takes a callback but ignores that callback.

### DIFF
--- a/test/express/server-specs.js
+++ b/test/express/server-specs.js
@@ -6,6 +6,7 @@ import request from 'request-promise';
 import chai from 'chai';
 import chaiAsPromised from 'chai-as-promised';
 import sinon from 'sinon';
+import B from 'bluebird';
 
 
 chai.should();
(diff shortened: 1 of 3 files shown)


test/express/server-specs.js
```diff
@@ -43,11 +44,16 @@ describe('server', () => {
       app.get('/error', () => {
         throw new Error('hahaha');
       });
+      app.get('/pause', async (req, res) => {
+        res.header['content-type'] = 'text/html';
+        await B.delay(3000);
```
jlipps: I like it, do we need this to be quite so long though? this is not an e2e test after all. could we get away with 500ms or 1s?

jlipps:
one comment then LGTM!

imurchie:
Cut it to 1s. Seems to remain reliable.

PR #56, opened 2016-06-28
Change name of xcuitest automation name

imurchie:
We are using `XCUITest` instead of `WebDriverAgent` as the `automationName`.


jlipps:
👍

PR #57, opened 2016-06-28
Make sure async is there, and add logging

imurchie:
Fix a problem where a command method is not `async`, and add a little logging for valid locator strategies.


jlipps:
👍

PR #58, opened 2016-07-01
Add ability to manage delegated drivers

imurchie:
We are getting in the situation where we need to have drivers that manage other drivers (e.g., `appium-xcuitest-driver` has an instance of `appium-safari-driver` that it delegates webbish stuff too). Rather than _ad hoc_ management of these drivers, let's move it into the base driver. 

At the moment only timeouts need to be handled.


jlipps:
👍

PR #61, opened 2016-07-06
Begin adding W3C route alternatives

imurchie:
We will, eventually, need to support the new spec, while also supporting the old versions until all the clients are updated. Keep track of that here.


jlipps:
👍

PR #63, opened 2016-07-11
Refactored mjsonwp with respect to UiAutomator2Driver functionality

sravanmedarapu:
Added condition in [`mjsonwp.js`](https://github.com/sravanmedarapu/appium-base-driver/blob/master/lib/mjsonwp/mjsonwp.js#L257) to return executed command status (i.e. [`driverRes`](https://github.com/sravanmedarapu/appium-base-driver/blob/master/lib/mjsonwp/mjsonwp.js#L267)) as response if driver is [`UiAutomator2Driver`](https://github.com/sravanmedarapu/appium-base-driver/blob/master/lib/mjsonwp/mjsonwp.js#L249) type.

Issue: https://github.com/appium/appium-uiautomator2-driver/issues/27

### DIFF
--- a/lib/mjsonwp/mjsonwp.js
+++ b/lib/mjsonwp/mjsonwp.js
@@ -1,10 +1,10 @@
 import _ from 'lodash';
 import { getLogger } from 'appium-logger';
 import { validators } from './validators';
-import { errors, isErrorType, MJSONWPError } from './errors';
+import { errors, isErrorType, MJSONWPError, errorFromCode } from './errors';
 import { METHOD_MAP, NO_SESSION_ID_COMMANDS } from './routes';
 import B from 'bluebird';
-
+import { util } from 'appium-support';
 
 const log = getLogger('MJSONWP');
 const JSONWP_SUCCESS_STATUS_CODE = 0;
@@ -243,6 +243,11 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
         log.debug(`Received response: ${_.truncate(JSON.stringify(driverRes), LOG_OBJ_LENGTH)}`);
         log.debug('But deleting session, so not returning');
         driverRes = null;
+      }
+
+      // if the status is not 0,  throw the appropriate error for status code.
+      if (util.hasValue(driverRes) && util.hasValue(driverRes.status) && parseInt(driverRes.status, 10) !== 0) {
+        throw errorFromCode(driverRes.status, driverRes.value);
       }
 
       // Response status should be the status set by the driver response.


imurchie:
Perhaps I'm not understanding the problem, but I _really_ don't like having this special code in here for a single driver.

First off, the response is from something that we have control over, so why not just make it conform to our standard?

Second, I would rather just see it be generalized. If I'm seeing the problem correctly, it is that the response has a status that is not `0`, but an error is not being thrown. This could happen from _anywhere_, not just the UiAutomator2 driver, no? So could we not just look at the status, and if it is not `0` throw the correct error, which would be caught like any other and the correct http status, etc., would be set? The `isUiAutomator2Driver` would not be needed.

```js
if (util.hasValue(driverRes) && util.hasValue(driverRes.status) && parseInt(driverRes.status, 10) !== 0) {
  throw errorFromCode(driverRes.status, driverRes.value);
}
```

imurchie:
Looks good! Thanks for this!

imurchie:
I've added a test, in #64.

imurchie:
Published in `appium-base-driver@2.0.12`.

lib/mjsonwp/mjsonwp.js
```diff
@@ -245,6 +245,11 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
         driverRes = null;
       }
 
+      // if the status is not 0,  throw the appropriate error for status code.
+      if (util.hasValue(driverRes) && util.hasValue(driverRes.status) && parseInt(driverRes.status, 10) !== 0) {
+        throw errorFromCode(driverRes.status, driverRes.value);
```
jlipps: I'm really confused why we'd want to throw something here rather than simply pass the (error) status back on to the client? Is it the case that we get MJSONWP objects from `driver.execute`? I thought we got actual responses

imurchie: The error handling code is in the exception handler [below](https://github.com/appium/appium-base-driver/blob/master/lib/mjsonwp/mjsonwp.js#L258-L268). Some drivers (e.g., WDA, and, at the moment, the UiAutomator2 driver) don't respond with exceptions, but set the status codes, so we would end up with a `200` http status with a Selenium error inside, which confuses the clients.

jlipps: hmm, to me this means we should fix the other drivers ASAP rather than add workarounds here. there should be two basic cases:

1) we call a driver command and it either returns a value or throws
2) we proxy to a JSONWP-based server, and the jsonwp-proxy code converts that into a return value or a throw

in no case should we try to do handle some other kind of response here IMO

sravanmedarapu: Sure, making the necessary changes to reflect the same

https://github.com/appium/appium-uiautomator2-server/pull/16

https://github.com/appium/appium-base-driver/pull/66

jlipps: I think we still need to settle on the overall design; the current design feels wrong to me. What do you think @imurchie

imurchie: I think moving this check into the proxy, and making any driver throw errors rather than just returning error codes, is good. I agree that catching it here muddies the waters.

PR #66, opened 2016-07-15
Implemented throwing ProxyRequestError

sravanmedarapu:
- ProxyRequestUnsuccessfulError will be thrown for all unsuccessful proxy requests, ProxyRequestUnsuccessfulError will contain the actual error route cause information in [`jsonwp`](https://github.com/sravanmedarapu/appium-base-driver/blob/5a66d7d9ec4772a9b88f9ba6545821b098f6c82f/lib/mjsonwp/errors.js#L316)

### DIFF
--- a/lib/mjsonwp/errors.js
+++ b/lib/mjsonwp/errors.js
@@ -1,6 +1,6 @@
 import ES6Error from 'es6-error';
 import _ from 'lodash';
-
+import { util } from 'appium-support';
 
 // base error class for all of our errors
 class MJSONWPError extends ES6Error {
@@ -308,6 +308,31 @@ class BadParametersError extends ES6Error {
   }
 }
 
+/**
+ * ProxyRequestError is a custom error and will be thrown up on unsuccessful proxy request and
+ * will contain information about the proxy failure.
+ * In case of ProxyRequestError should fetch the actual error by calling `getActualError()`
+ * for proxy failure to generate the client response.
+ */
+class ProxyRequestError extends ES6Error {
+  constructor (err, jsonwp) {
+    let message = `Proxy request unsuccessful. ${util.hasValue(jsonwp) ? jsonwp.value : ''}`;
+    super(err || message);
+    if (typeof jsonwp === 'string') {
+      this.jsonwp = JSON.parse(jsonwp);
+    } else {
+      this.jsonwp = jsonwp;
+    }
+  }
+
+  getActualError () {
+    if (util.hasValue(this.jsonwp) && util.hasValue(this.jsonwp.status) && util.hasValue(this.jsonwp.value)) {
+      //returns actual error cause for request failure based on `jsonwp.status`
+      return errorFromCode(this.jsonwp.status, this.jsonwp.value);
+    }
+    return new UnknownError(this.message);
+  }
+}
 // map of error class name to error class
 const errors = {NotYetImplementedError,
                 NotImplementedError,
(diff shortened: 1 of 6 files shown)


lib/mjsonwp/errors.js
```diff
@@ -308,6 +308,26 @@ class BadParametersError extends ES6Error {
   }
 }
 
+class ProxyRequestUnsuccessfulError extends ES6Error {
```
imurchie: I don't think "Unsuccessful" is necessary here. An error, by definition, marks something unsuccessful.

I'd rather see `ProxyRequestError`.

lib/mjsonwp/errors.js
```diff
@@ -308,6 +308,26 @@ class BadParametersError extends ES6Error {
...
+  constructor(err, jsonwp) {
+    super(err || "Proxy Request Unsuccessful" + jsonwp.value);
```
imurchie: This should be an interpolated string: 

```js
super(err || `Proxy Request Unsuccessful ${jsonwp.value}`);
```

imurchie:
Looks good!

Could you add a small comment to the error class definition, describing why it is there (it is the only error that does not map to a Selenium error).

imurchie:
And, finally, a unit test or two would be good. One in the [error specs](https://github.com/appium/appium-base-driver/blob/master/test/mjsonwp/errors-specs.js) and at least one in the [mjsonwp specs](https://github.com/appium/appium-base-driver/blob/master/test/mjsonwp/mjsonwp-specs.js) (modeled after [this](https://github.com/appium/appium-base-driver/blob/master/test/mjsonwp/mjsonwp-specs.js#L40-L51)).

sravanmedarapu:
Sure :+1:

lib/mjsonwp/mjsonwp.js
```diff
@@ -257,7 +257,9 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
                `result: ${_.truncate(JSON.stringify(driverRes), LOG_OBJ_LENGTH)}`);
     } catch (err) {
       let actualErr = err;
-      if (!(isErrorType(err, MJSONWPError) ||
+      if (err instanceof errors.ProxyRequestUnsuccessfulError) {
```
jlipps: we actually can't use `instanceof` checks here because they will fail across package versions.

i.e., an instance of this error class for v1.2.3 or base-driver is not the same as one in v1.2.4. this is why we made the `isErrorType` function instead.

sravanmedarapu:
Added specs and updated using `isErrorType` instead of `instanceof` as per @jlipps comment.

lib/mjsonwp/errors.js
```diff
@@ -308,6 +308,28 @@ class BadParametersError extends ES6Error {
   }
 }
 
+class ProxyRequestError extends ES6Error {
+  constructor(err, jsonwp) {
+    var message = `Proxy Request Unsuccessful. `;
```
imurchie: This should be `let`. It also does not need to be in the special backtick-enclosed string. Or it can be rewritten as

```js
let message = `Proxy request unsuccessful. ${util.hasValue(jsonwp} ? jsonwp.value : ''}`;
```

imurchie:
LGTM

imurchie:
This looks good to me. Is it ready to go?

sravanmedarapu:
Yeah, ready to go.

imurchie:
Published in `appium-base-driver@2.0.14`.

PR #67, opened 2016-07-17
added rotation routes

rafael-chavez:
@jlipps @imurchie 

I know this involves changing the hash for the gulp tests. But before thats done want to make sure this pull is ok. 

Here is the spec for this [route](https://github.com/SeleniumHQ/mobile-spec/blob/master/spec-draft.md#device-orientation)

### DIFF
--- a/lib/mjsonwp/routes.js
+++ b/lib/mjsonwp/routes.js
@@ -175,6 +175,10 @@ const METHOD_MAP = {
     GET: {command: 'getOrientation'},
     POST: {command: 'setOrientation', payloadParams: {required: ['orientation']}}
   },
+  '/wd/hub/session/:sessionId/rotation': {
+    GET: {command: 'getRotation'},
+    POST: {command: 'setRotation', payloadParams: {required: ['x', 'y', 'z']}}
+  },
   '/wd/hub/session/:sessionId/moveto': {
     POST: {command: 'moveTo', payloadParams: {optional: ['element', 'xoffset', 'yoffset']}}
   },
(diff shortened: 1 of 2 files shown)


imurchie:
The spec seems to say that it should accept a `DeviceRotation` object, but their examples are sending `x`, `y`, and `z` values (https://github.com/SeleniumHQ/mobile-spec/blob/master/spec-draft.md#devicerotation). Not sure which is to be taken as correct.

jlipps:
`DeviceRotation` is the "name" of the object type, similarly to `NetworkConnection` objects. that name doesn't actually need to show up in the JSONWP though.

imurchie:
Ah. Ok. That needs to be reflected here, then.

rafael-chavez:
Are we saying the post part should change like this?

```
POST: {command: 'setRotation', payloadParams: {required: ['rotation']}}
```
 to :
```
POST: { command: 'setRotation', payloadParams: { unwrap: 'parameters', required: ['type'] } }
```

imurchie:
No. The spec says the payload should be something like

```json
{"name": "rotation", "parameters": {"x": 0, "y": 0, "z": 0}}
```

So in the mjsonwp implementation here, it should be something like

```js
payloadParams: {unwrap: 'parameters', required: ['x', 'y', 'z']}
```

rafael-chavez:
updated mobile spec [here](https://github.com/SeleniumHQ/mobile-spec/pull/5)

imurchie:
Looks good to me.

imurchie:
And thanks for fixing the spec!

rafael-chavez:
@imurchie should i update the expected test hash?

imurchie:
Yes. Otherwise the tests will fail and we can't merge.

imurchie:
Thanks!

PR #68, opened 2016-07-29
Add homescreen endpoint

ericwengelking:
This is a new endpoint provided by https://github.com/facebook/WebDriverAgent

It can be used to simulate tapping on the home button on an iOS device

```
curl -X POST $JSON_HEADER -d "" $DEVICE_URL/homescreen
```

### DIFF
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -26,7 +26,7 @@ class JWProxy {
   }
 
   endpointRequiresSessionId (endpoint) {
-    return !_.includes(['/session', '/sessions', '/status'], endpoint);
+    return !_.includes(['/session', '/sessions', '/status', '/homescreen'], endpoint);
   }
 
   getUrlForProxy (url) {
@@ -34,7 +34,7 @@ class JWProxy {
       url = '/';
     }
     const proxyBase = `${this.scheme}://${this.server}:${this.port}${this.base}`;
-    const endpointRe = '(/(session|status))';
+    const endpointRe = '(/(session|status|homescreen))';
     let remainingUrl = '';
     if (/^http/.test(url)) {
       const first = (new RegExp(`(https?://.+)${endpointRe}`)).exec(url);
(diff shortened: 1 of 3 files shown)


lib/jsonwp-proxy/proxy.js
```diff
@@ -26,15 +26,15 @@ class JWProxy {
   }
 
   endpointRequiresSessionId (endpoint) {
-    return !_.includes(['/session', '/sessions', '/status'], endpoint);
+    return !_.includes(['/session', '/sessions', '/status', '/homescreen'], endpoint);
```
imurchie: Why is this a sessionless command? I understand that it is in Facebook's [WDA](https://github.com/facebook/WebDriverAgent/blob/2031c1ccfaf7242cf48e6086fe58a9f46ed3c3ed/WebDriverAgentLib/Commands/FBCustomCommands.m#L32), but I'm wondering of the actual use case.

ericwengelking: It doesn't require a sesion.  With Facebook's WDA you don't need a session to execute this command.  Also with Facebook's WDA you can automate part of simulator outside of your application.  

I could send a pull request to Facebook's WDA and see if we can also add `homescreen` as a session command if that would work better.

imurchie: It can be proxied to WDA without a session. I'm just curious as to why, in Appium, it should be possible to do it.

It will require significant energy to get it to work in Appium, since Appium does not have a connection to the device without a session. All the current sessionless commands do nothing to the device.

ericwengelking: I think people will only use this once an Appium session has already been started.

imurchie: Alright. It should be removed from this list, then, and just added as a normal endpoint.

tateexon:
@ericwengelking Any plans to move forward with this or is it dead in the water?

lib/jsonwp-proxy/proxy.js
jlipps: @ericwengelking think this change still needs to be made?

jlipps:
closing for now.

PR #69, opened 2016-08-10
Allow guinea pig delay to be in query

imurchie:
Allow the delay to be specified in either the body or the query params.

### DIFF
--- a/lib/express/static.js
+++ b/lib/express/static.js
@@ -13,7 +13,7 @@ if (_.isNull(path.resolve(__dirname).match(/build[\/\\]lib[\/\\]express$/))) {
 }
 
 async function guineaPigTemplate (req, res, page) {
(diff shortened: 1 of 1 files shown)


lib/express/static.js
```diff
@@ -13,7 +13,9 @@ if (_.isNull(path.resolve(__dirname).match(/build[\/\\]lib[\/\\]express$/))) {
 }
 
 async function guineaPigTemplate (req, res, page) {
-  let delay = req.params.delay ? parseInt(req.params.delay, 10) : 0;
+  let delay = req.params.delay ? parseInt(req.params.delay, 10)
```
jlipps: idea:

```
let delay = parseInt(req.params.delay || req.query.delay || '0', 10);
```

jlipps:
👍

lib/express/static.js
imurchie: Nice!

PR #70, opened 2016-08-12
adding locator and keysTosend as optional Parameters

yaphatak:


imurchie:
Can these not be added as optional parameters to the endpoints themselves?

In the manner of https://github.com/appium/appium-base-driver/blob/master/lib/mjsonwp/routes.js#L12

The other optional parameters in the `checkParams` function are things that get sent _all the time_ by certain clients, rather than just for particular commands.

yaphatak:
Yes. Let me do that. Thanks

imurchie:
One further note. There is a test that is used to make sure that changes to the protocol are intentional. Essentially it takes a hash of the protocol and compares it to what is expected.

When you change routes, and some other parts of the protocol package, you need to update that test. https://github.com/appium/appium-base-driver/blob/master/test/mjsonwp/routes-specs.js#L41

The easiest way is to, when you are ready to check it in, run `gulp once` and use the failing notification to find the hash for the newly changed protocol.

imurchie:
This looks good to me!

But, is there no way for the C# client to _not_ send these?

jlipps:
Yes, what is the real point of adding these? Are these part of the spec? If it is a client that is sending the wrong thing, we should probably fix the client...?

yaphatak:
@jlipps , @imurchie Thanks for quick response. Actually this got fixed in latest version (2.51). Our team was using 2.45. Let me close this. We dont need this change

PR #71, opened 2016-08-12
Add web context check to validate locators

imurchie:
Add check for web context within the generic find methods. This way sub-classes can implement a `isWebContext` method in order to get the correct checks, rather than needing to implement their own `validateLocatorStrategy` method that is boilerplate.


jlipps:
if we're maintaining this as state anyway, why not just have `validateLocatorStrategies` check `this.isWebContext()`?

imurchie:
We totally could. Unless we put it like this for a reason, and I'm not sure about that.

PR #72, opened 2016-09-03
Removing _truncate for requests for better debugging

rkavalap:
Removing _truncate for requests for better debugging
All the proxied requests are logged truncated in log files. It's hard to identify what was the original request which failed.

### DIFF
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -8,9 +8,8 @@ import { errors } from '../mjsonwp/errors';
 
 
 const log = getLogger('JSONWP Proxy');
-// set a reasonable log length for proxy output, so we can see what is being sent.
-// in the future maybe make this configurable?
-const LOG_LENGTH = 1000;
+// TODO: Make this value configurable as a server side capability
+const LOG_OBJ_LENGTH = 1024; // MAX LENGTH Logged to file / console
 
 class JWProxy {
   constructor (opts = {}) {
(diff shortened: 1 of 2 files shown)


imurchie:
I don't think we should do this without some sort of checks on what the request/response is. What if the user is pushing or pulling a 10Mb file?

rkavalap:
@imurchie Agreed. Instead of 150 / 200 characters which was previous, 
what do you say about (512 * 1024) Bytes ? This is much smaller, should cover 99% of the requests. The only requests which would be truncated i expect would be something like push / pull file, /log api of selenium web driver. If that's fine with you let me know and ill update the PR.

rkavalap:
@imurchie  Considering the statement is a log.debug, the user can set the --log-level to INFO if they do not want to see the log message. It's almost like saying you are turning on verbose mode. The only caveat being, default log level is DEBUG in appium. The reason I chose 512KB was, chromedriver itself limits to 256KB / 512KB when i was investigating an issue of invalid json for the /log api. I am not a big fan of server side flag for this. But it definitely brings a bigger discussion of which messages should be put in log.debug vs log.info etc. Considering there are way too many packages in appium it's hard to track down every log message to make the distinction

imurchie:
Sorry for the delay with this. I still think that we don't want to dump so much in by default, even at the debug level. This could be behind a server argument or desired capability so that people could turn it on if they wanted.

jlipps:
@rkavalap tell you what. how about we set it to 1k by default, and then create a capability called `fullResponseLogging` that when true, doesn't truncate response logs at all?

rkavalap:
@jlipps fair enough. I just rebased and pushed changes. Realized @imurchie  already made a change recently for one of the proxy.

jlipps:
seems OK to me now

lib/jsonwp-proxy/proxy.js
```diff
@@ -108,13 +109,15 @@ class JWProxy {
       reqOpts.json = null;
     }
 
-    log.info(`Proxying [${method} ${url || "/"}] to [${method} ${newUrl}] ` +
-             (body ? `with body: ${_.truncate(JSON.stringify(body), {length: LOG_LENGTH})}` : 'with no body'));
+    log.debug(`Proxying [${method} ${url || "/"}] to [${method} ${newUrl}] ` +
+             (body ? `with body: ${_.truncate(JSON.stringify(body), LOG_OBJ_LENGTH)}` : 'with no body'));
```
imurchie: This isn't how `_.truncate` works. The second argument has to be an object with the length as `length` property

```js
_.truncate(JSON.stringify(body), {length: LOG_OBJ_LENGTH})
```

rkavalap: Bad merges. Ill push another change for this.

rkavalap: @imurchie  Fixed now.

imurchie:
Thanks! Published in `appium-base-driver@2.0.25`

PR #73, opened 2016-09-09
adding rect routes

wfanforever:
@jlipps @imurchie @rafael-chavez 

Adding a new routes to match wda: 
https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/remote/DriverCommand.java#L83

WDA:
https://github.com/facebook/WebDriverAgent/blob/master/WebDriverAgentLib/Commands/FBElementCommands.m#L46

new spec:
https://w3c.github.io/webdriver/webdriver-spec.html#get-element-rect

### DIFF
--- a/lib/mjsonwp/routes.js
+++ b/lib/mjsonwp/routes.js
@@ -421,6 +421,10 @@ const METHOD_MAP = {
   '/wd/hub/session/:sessionId/alert/dismiss': {
     POST: {command: 'postDismissAlert'}
   },
+  // https://w3c.github.io/webdriver/webdriver-spec.html#get-element-rect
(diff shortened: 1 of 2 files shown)


imurchie:
This route should go down in the bottom section (https://github.com/appium/appium-base-driver/blob/master/lib/mjsonwp/routes.js#L396-L423) where we have begun to add W3C spec routes.

wfanforever:
@imurchie I have updated the PR to move it to the bottom. :)

jlipps:
👍

imurchie:
Excellent! Thanks for this!

PR #75, opened 2016-09-13
Copy from WindowsNetworkShare

yaphatak:
@imurchie : Look at this whenever you get time. we want to add support to copy app from windows network share. I know this is not perfect implementation. So, if you could help with right approach, it will be awesome. Also, I would like to add unit test but I am not sure how that could run with CI.

### DIFF
--- a/lib/basedriver/helpers.js
+++ b/lib/basedriver/helpers.js
@@ -12,7 +12,7 @@ import request from 'request';
 
 const ZIP_EXTS = ['.zip', '.ipa'];
 
-async function configureApp (app, appExt) {
+async function configureApp (app, appExt, mountRoot="Volumes", windowsShareUserName="", windowsSharePassword="") {
   if (_.isNull(app) || _.isUndefined(app)) {
     // immediately shortcircuit if not given an app
     return;
@@ -20,6 +20,12 @@ async function configureApp (app, appExt) {
 
   let newApp = null;
   let shouldUnzipApp = _.includes(ZIP_EXTS, path.extname(app));
+
+  // check if we're copying from a windows network share
+  if (_.startsWith(app, "\\")) {
+    app = await copyFromWindowsNetworkShare(app, appExt, mountRoot, windowsShareUserName, windowsSharePassword);
+  }
+
   if ((app || '').substring(0, 4).toLowerCase() === 'http') {
     logger.info(`Using downloadable app '${app}'`);
     newApp = await downloadApp(app, shouldUnzipApp ? '.zip' : appExt);
@@ -31,7 +37,7 @@ async function configureApp (app, appExt) {
       newApp = await copyLocalZip(app);
     }
   }
-
+  
   if (shouldUnzipApp) {
     newApp = await unzipApp(newApp, appExt);
     logger.info(`Unzipped local app to '${newApp}'`);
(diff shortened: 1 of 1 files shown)


lib/basedriver/helpers.js
```diff
@@ -12,14 +12,20 @@ import request from 'request';
 
 const ZIP_EXTS = ['.zip', '.ipa'];
 
-async function configureApp (app, appExt) {
+async function configureApp (app, appExt, mountroot="Volumes", windowsShareUserName="", windwosSharePassword="") {
```
imurchie: `windowsSharePassword`

lib/basedriver/helpers.js
```diff
@@ -202,6 +208,66 @@ async function testZipArchive (zipPath) {
...
+  }
+  return app;
+}
+
+async function copyLocallyFromWindowsShare (app, appExt) {
+  let fileInfo = await tempDir.open({prefix: 'appium-app', suffix: appExt});
+  return await copyFile(app, fileInfo.path);	
+}
+
+async function copyFile (source, target) {
```
imurchie: Does `fs.copyFile` from [appium-support](https://github.com/appium/appium-support/blob/master/lib/fs.js#L34) not work?

lib/basedriver/helpers.js
```diff
@@ -202,6 +208,66 @@ async function testZipArchive (zipPath) {
...
+      await exec('mount', mountArgs, { maxBuffer: 524288 });
+    } catch (err) {
+      logger.errorAndThrow(`Error mounting: ${err.message}`);
+    }
+  };
+  if (await fs.exists(mountPath)) {
+    if (await fs.exists(app)) {
+      return app;
+    } else {
+      let umountArgs = [`${mountPath}`];
```
imurchie: No need to have `mountPath` in a string interpolation, since it is already a string. `let unmountArgs = [mountPath];` will suffice.

lib/basedriver/helpers.js
```diff
@@ -202,6 +208,66 @@ async function testZipArchive (zipPath) {
...
+      try {
+        await exec('umount', umountArgs, { maxBuffer: 524288 });
+      } catch (err) {
+        logger.error(`Error Unmounting :${err.message}`);
+      }
+      await fs.rimraf(mountroot);
+      await mountNetworkShare();
```
imurchie: I would move both these out, so the `if (await fs.exists(mountPath)) {}` has no `else` (the other else is unnecessary too, because of the return).

 ```js
  if (await fs.exists(mountPath)) {
    if (await fs.exists(app)) {
      return app;
    }
    let umountArgs = [mountPath];
    try {
      await exec('umount', umountArgs, { maxBuffer: 524288 });
    } catch (err) {
      logger.error(`Error Unmounting: ${err.message}`);
    }
    await fs.rimraf(mountroot);
  }

  await mountNetworkShare();
  return app;
```

File: lib/basedriver/helpers.js (hunk 2 above)
Comment (yaphatak) [inline lib/basedriver/helpers.js]: I didn't realize that its there I used they way `copyLocalZip` was doing.

File: lib/basedriver/helpers.js (hunk 1 above)
Comment (yaphatak) [inline lib/basedriver/helpers.js]: thanks

File: lib/basedriver/helpers.js (hunk 4 above)
Comment (yaphatak) [inline lib/basedriver/helpers.js]: I didn't realize I was doing this stupidity. I translated my earlier code to ES6 yesterday. Thanks.

**Comment by imurchie** [comment]
This looks good to me now! Can you squash it into a single commit?

File: lib/basedriver/helpers.js [hunk 5]
```diff
@@ -12,14 +12,20 @@ import request from 'request';
 
 const ZIP_EXTS = ['.zip', '.ipa'];
 
-async function configureApp (app, appExt) {
+async function configureApp (app, appExt, mountroot="Volumes", windowsShareUserName="", windowsSharePassword="") {
```
jlipps: small thing, but I'd prefer `mountRoot` to `mountroot`

lib/basedriver/helpers.js
```diff
@@ -12,14 +12,20 @@ import request from 'request';
...
   if (_.isNull(app) || _.isUndefined(app)) {
     // immediately shortcircuit if not given an app
     return;
   }
 
   let newApp = null;
   let shouldUnzipApp = _.includes(ZIP_EXTS, path.extname(app));
+  let shouldCopyFromWindowsNetworkShare = _.startsWith(app, "\\");
+
+  if (shouldCopyFromWindowsNetworkShare) {
```
jlipps: again, a small thing. Maybe this is more readable, but since we're only using this variable in one place, we could more succinctly write this as:

```js
// check if we're copying from a windows network share
if (_.startsWith(app, "\\")) {
```

lib/basedriver/helpers.js
```diff
@@ -202,6 +208,53 @@ async function testZipArchive (zipPath) {
...
+  let rootFolder = pathSplit[3];
+  app = app.replace(/\\/g, "/");
+  app = app.replace(`/${networkShare}`, mountroot);
+  let mountPath = `/${mountroot}/${rootFolder}`;
+
+  let mountNetworkShare = async function () {
+    await fs.mkdir(mountPath);
+    let mountArgs = [`-t`, `smbfs`, `${windowsUserName}:${windowsPassword}@${networkShare}/${rootFolder}`, mountPath];
+    try {
+      await exec('mount', mountArgs, { maxBuffer: 524288 });
```
jlipps: we don't need to specify the `maxBuffer` option since we're using `teen_process`'s exec. it will have no effect.

lib/basedriver/helpers.js
```diff
@@ -202,6 +208,53 @@ async function testZipArchive (zipPath) {
...
+    }
+  };
+
+  if (await fs.exists(mountPath)) {
+    if (await fs.exists(app)) {
+      return app;
+    } 
+    let umountArgs = [mountPath];
+    try {
+      await exec('umount', umountArgs, { maxBuffer: 524288 });
```
jlipps: same here w.r.t. maxBuffer

yaphatak:
Thanks @jlipps for feedback. I will have those changes made. I am glad more people are looking into PRs, it helps.

jlipps:
@yaphatak thanks, please just squash again and we should be good to go.

yaphatak:
@imurchie : are you merging this?

imurchie:
Totally! I had forgotten about this.

imurchie:
Published in `appium-base-driver@2.0.17`

PR #78, opened 2016-09-30
Use keep-alive connection for proxying

imurchie:
xcuitest real devices work significantly better if the connection is not closed on each request. Other systems seem to work fine in keep-alive mode, too.


imurchie:
Published in `appium-base-driver@2.0.19`.

PR #82, opened 2016-10-26
add apis to read the performance data

heeseon:
the appium already has a way to gather the performance data by log.
but If the resulting data format is set to a log file, the user should conduct another parsing to show the data on the screen.
also I think user wants to decide the moment to read the data and associate the performance data and other information.
so I’d like to add apis to gather the performance data.

### DIFF
--- a/docs/mjsonwp/protocol-methods.md
+++ b/docs/mjsonwp/protocol-methods.md
@@ -109,6 +109,8 @@ POST        | `/wd/hub/session/{sessionId}/appium/device/shake`                 
 POST        | `/wd/hub/session/{sessionId}/appium/device/lock`                       | Lock the device.
 POST        | `/wd/hub/session/{sessionId}/appium/device/unlock`                     | Unlock the device.
 POST        | `/wd/hub/session/{sessionId}/appium/device/is_locked`                  | Check whether the device is locked or not.
+POST        | `/wd/hub/session/{sessionId}/appium/performanceData/types`             | returns the information types of the system state which is supported to read as like cpu, memory, network traffic, and battery.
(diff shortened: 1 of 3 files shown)


imurchie:
Two things:

1. These will need to be within the appium extension (i.e., `/wd/hub/session/:sessionId/appium`) since they are not part of the Selenium spec.

2. I do not understand the issue at stake here, from your explanation. What will the methods on the server side do, when they get these requests? I think perhaps a concrete implementation of this for one of the drivers might be a better first step, before adding the routes to the protocol.

heeseon:
Um.. the user can gather the performance data with the log for web and iOS. But the log file format is fixed. so the user can't change the format. 
Now, I need to gather the performance data for the android. 
But now, user can't gather the performance data for android because the appium-android-driver only supports the logcat log.
So, I thought I have to add the log type of the performance.
But, I found some restriction for gathering the performance data as the log file.
For example, I'd like to connect the performance data with test case number or some function is called. 
But it is not impossible for now because current implementation is just read the performance data periodically.
So, I decide to define the apis to read the performance data when user wants to read it. so uesr can process the result with other information(for example test number or the specific function is called).
and also the user does not need to understand the result file format and parse it to process the result.
You still think that it does not need?

heeseon:
If you think it does not need, then I will define the api within the appium extension.

heeseon:
I moved the api to appium extension.

heeseon:
I moved the apis to appium extension

imurchie:
There is a conflict here, so we cannot merge it as-is. Can you see what that is, and fix?

heeseon:
I can't check the conflicting file. 
In my PC, build success below command.
npm test

or 

gulp

is it ok that I close this pull request and make another branch and send new pull request?

imurchie:
If that is easier for you, sure!

The conflict is with the hash in `route-specs.js`. It was changed by another commit too.

PR #84, opened 2016-10-31
Make use of os.tmpdir instead of os.tmpDir

imurchie:
`os.tmpDir` is deprecated in Node 7.


imurchie:
Once this gets in Appium starts cleanly in Node 7!

jlipps:
\o/

imurchie:
Published in `appium-base-driver@2.0.23`

Issue #79, opened 2016-10-03
Unzipping relies on file extension


sa-spag:
According to [the source code](https://github.com/appium/appium-base-driver/blob/7af4333f9e3005d9c6d4feec274cca8af773d654/lib/basedriver/helpers.js#L22) it is the file extension which determines wether or not the file should be unzipped.

**The problem:** I cannot use URLs without extensions.
**The trick:** Using fake URL parameters, such as `https://my.url/without/extension?fake=.zip`.
**The solution(s):** Using a module which does not rely on the file extension (e.g. [`file-type`](https://www.npmjs.com/package/file-type) or allow the file format to be passed as a desired capability?

sa-spag:
Thanks @midopa!

PR #86, opened 2016-11-05
Handle zip files from URLs

midopa:
Addresses https://github.com/appium/appium-base-driver/issues/79

Sometimes you can't tell the file type from a URL. Like URL minimizers or if there are parameters, etc. So just check the content type of the response. If it's a zip file, ensure it gets unzipped in subsequent logic. Otherwise, assume it's an app file.

Works on:
- URL pointing to an APK file
- URL pointing to a ZIP file
- URL pointing to a ZIP file but with parameters in the URL

Note: The `downloadApp` and `downloadFile` functions are exported in helpers. Their signatures have changed. But I did a full grep on all the code from a fresh Appium node install (including node_modules) and saw the only users were inside `helpers.js` so I don't think this is would break anyone.

### DIFF
--- a/lib/basedriver/helpers.js
+++ b/lib/basedriver/helpers.js
@@ -11,6 +11,7 @@ import request from 'request';
 
 
 const ZIP_EXTS = ['.zip', '.ipa'];
+const ZIP_MIME_TYPE = 'application/zip';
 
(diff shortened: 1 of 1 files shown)


imurchie:
It's not used outside this one file (it doesn't need to be exported, even, and is not exported by the overall package).

imurchie:
Published in `appium-base-driver@2.0.24`.

PR #88, opened 2016-11-12
Fix handling non-ZIP files

midopa:
APK files were not being installed because they were being written to a file without an "apk" extension. So added back the appExt suffix logic. If it's given and the downloaded file is not a ZIP file, then everything behaves normally. If the downloaded file is a ZIP file, then regardless of what we thought initially, it will be unzipped and placed into another temp location.


midopa:
Added e2e tests.

Is `appExt` is really needed? It seems the drivers should care about that. These utils just download and extract files. It can simplify the code and the tests. But it'd require coordination across all the drivers that use this (Android, XCUITest, iOS, Selenium, UiAutomator2, maybe more).

midopa:
Ping on this one?

imurchie:
I'm good with this. Thanks!

imurchie:
As for the `appExt`, that would require more thought. In the meantime, this is good to have in!

imurchie:
Published in `appium-base-driver@2.0.26`.

PR #89, opened 2016-11-12
Add routes and docs for device commands

midopa:
For: https://github.com/appium/appium-android-driver/pull/192

It seems a lot of clients have to be updated to support these out of the box. I wasn't able to do `driver.execute('display_density')` in Python because the Appium client uses the Selenium client which does a look up in a table of predefined commands. I assume clients for all the other languages are the same.

Work around to get functionality is to build a URL and make a request without using a client, which is how I tested this.


imurchie:
The unit tests have one that is designed to catch unexpected changes to the protocol. Essentially, the hash of the protocol is checked. You'll need to update that test with the new hash.

That test ought to have been run in a pre-commit hook. I'm surprised the commit could happen with the test failing.

midopa:
Updated.

The pre-commit hook doesn't seem to be setup in my local clone.

I tried cloning it again (in case the setup stuff was added recently), but the pre-commit hooks still don't fire.

Oddly, this just works for the Android driver I have cloned?

imurchie:
It should be installed during the `npm install` phase for the repo. Weird.

imurchie:
LGTM. Thanks!

midopa:
FWIW, I removed `node_modules` on my local clone, re-ran `npm install`, and tried to do a commit. The pre-commit hook now fires. So I guess I cloned and installed this a long time ago and since then npm was just doing incremental builds?

PR #90, opened 2016-11-15
Support `pageLoadStrategy` capability (now supported by chromedriver).

tonyhignett:
Chromedriver now supports the `pageLoadStrategy` capability (though they're not desperately keen to tell anyone; it's not in the docs: https://sites.google.com/a/chromium.org/chromedriver/capabilities). It's specified here: https://www.w3.org/TR/webdriver/#dfn-page-load-strategy.

So we'd like to be able to pass it through appium, which requires it to be added to the validation constraints.

### DIFF
--- a/lib/basedriver/desired-caps.js
+++ b/lib/basedriver/desired-caps.js
@@ -58,7 +58,10 @@ let desiredCapabilityConstraints = {
   },
   locale: {
     isString: true
-  }
+  },
+  pageLoadStrategy: {
+    isString: true
+  },
 };
 
 validator.validators.isString = function (value) {


imurchie:
This is only valid for Android, correct? So it should go in https://github.com/appium/appium-android-driver/blob/master/lib/desired-caps.js

tonyhignett:
OK, I can do that. It seems unnecessary to duplicate the validation performed by the individual drivers, though. (I was surprised I had to add it to appium at all, since appium is not doing anything with it.)

imurchie:
It is not duplicated, it is just where it needs to be. The base driver only has the generic capabilities used throughout. The individual drivers keep track of what is particular to that system.

You have to add it, because Appium doesn't want random things passed in. It makes debugging more difficult than it already is.

PR #95, opened 2016-12-20
Do not keep adding sessionId and id to optional parameters

imurchie:
Check before adding `sessionId` and `id` to the optional parameters. Otherwise, after a number of sessions, we end up with things like:

```
dbug MJSONWP Calling XCUITestDriver.createSession() with args: [{"safariIgnoreFraudWarning":false,"safariInitialUrl":"http://0.0.0.0:4994/test/guinea-pig","safariAllowPopups":true,"nativeWebTap":true,"browserName":"Safari","platformName":"iOS","platformVersion":"9.3","deviceName":"iPhone 6","automationName":"XCUITest","noReset":true},null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null]
```

### DIFF
--- a/test/mjsonwp/routes-specs.js
+++ b/test/mjsonwp/routes-specs.js
@@ -38,7 +38,7 @@ describe('MJSONWP', () => {
       }
       var hash = shasum.digest('hex').substring(0, 8);
       // Modify the hash whenever the protocol has intentionally been modified.
-      hash.should.equal('5b035b99');
+      hash.should.equal('f5aec588');
     });
   });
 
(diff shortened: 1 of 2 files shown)


test/mjsonwp/routes-specs.js
```diff
@@ -38,7 +38,7 @@ describe('MJSONWP', () => {
       }
       var hash = shasum.digest('hex').substring(0, 8);
       // Modify the hash whenever the protocol has intentionally been modified.
-      hash.should.equal('5b035b99');
+      hash.should.equal('f5aec588');
```
jlipps: was this not changed before? don't see any routes edits in this PR

PR #99, opened 2017-01-04
Add configurable request timeout to proxy

imurchie:
There are use cases for having the connection open longer. See https://github.com/appium/appium/issues/7643


imurchie:
Yes. But the proxy management is handled at the driver level, so it has to happen there. Alas.

jlipps:
makes sense

PR #100, opened 2017-01-13
fixing copyFromWindowsShare on Windows 

yaphatak:
fixing copyFromWindowsShare on Windows as fs.copy really doesn't do any copy on windows


imurchie:
Interesting. I think this would, then, be better placed in `appium-support` so that on Windows `fs.copyFile` does what it should. Otherwise we need to copy this code around to everywhere that something needs to be copied.

yaphatak:
Okay then i will close this one for now?

PR #101, opened 2017-01-21
Default to application/json when no Content-Type is supplied

maxhawkins:
The webdriver spec specifies that requests where the method is POST
should be parsed as JSON (see https://www.w3.org/TR/webdriver/#processing-model step 6).

Appium is too restrictive. It only parses as JSON if the application/json
Content-Type is provided, making it incompatible with some Webdriver clients
that follow the spec but don't send a Content-Type header.

This commit makes Appium less restrictive by defaulting to the application/json
Content-Type if none is provided.

### DIFF
--- a/lib/express/middleware.js
+++ b/lib/express/middleware.js
@@ -30,6 +30,13 @@ function fixPythonContentType (req, res, next) {
   next();
 }
 
+function defaultToJSONContentType (req, res, next) {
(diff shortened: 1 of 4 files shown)


jlipps:
This is a reasonable change, and good implementation. And thanks for tests! 👍

jlipps:
published in `appium-base-driver@2.1.5`

PR #104, opened 2017-02-10
Add METHOD_MAP to public exports, so it can be used by other modules

mykola-mokhnach:
subj

### DIFF
--- a/lib/mjsonwp/routes.js
+++ b/lib/mjsonwp/routes.js
@@ -455,7 +455,56 @@ for (let v of _.values(METHOD_MAP)) {
   }
 }
 
+const RE_ESCAPE = /[\-\[\]{}()+?.,\\\^$|#\s]/g;
+const RE_PARAM = /([:*])(\w+)/g;
+
+class Route {
+  constructor (route) {
+    this.paramNames = [];
+
+    let reStr = route.replace(RE_ESCAPE, "\\$&");
+    reStr = reStr.replace(RE_PARAM, (_, mode, name) => {
+      this.paramNames.push(name);
+      return mode === ":" ? "([^/]*)" : "(.*)";
+    });
+    this.routeRegexp = new RegExp(`^${reStr}$`);
+  }
+
+  parse (url) {
+    let matches = url.match(this.routeRegexp);
+    if (!matches) return;
+    let i = 0;
+    let params = {};
+    while (i < this.paramNames.length) {
+      const paramName = this.paramNames[i++];
+      params[paramName] = matches[i];
+    }
+    return params;
+  }
+}
+
+function routeToCommandName (endpoint, method) {
+  let dstRoute = null;
+  const actualEndpoint = _.startsWith(endpoint, '/') ? endpoint : `/${endpoint}`;
+  for (let currentRoute of _.keys(METHOD_MAP)) {
+    const route = new Route(currentRoute);
+    // we don't care about the actual session id for matching
+    if (route.parse(`/wd/hub/session/ignored-session-id${actualEndpoint}`) ||
+        route.parse(`/wd/hub${actualEndpoint}`) || route.parse(actualEndpoint)) {
+      dstRoute = currentRoute;
+      break;
+    }
+  }
+  if (!dstRoute) return;
+
+  const methods = _.get(METHOD_MAP, dstRoute);
+  if (_.has(methods, method)) {
+    const dstMethod = _.get(methods, method);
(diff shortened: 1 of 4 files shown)


imurchie:
Is there any chance that this will be used in multiple places? If it is, and for the same purpose (reverse method lookup) it might be better (rather than having to do the reversal in multiple places) to either (a) reverse here and export that, or (b) add a method here to look up the method name for a given driver command function name.

mykola-mokhnach:
@imurchie I can borrow the lookup implementation from https://github.com/appium/appium-xcuitest-driver/pull/360

mykola-mokhnach:
done. Please take a look

lib/mjsonwp/routes.js
```diff
@@ -455,7 +455,56 @@ for (let v of _.values(METHOD_MAP)) {
...
+  constructor (route) {
+    this.paramNames = [];
+    this.routeRegexp = route;
+
+    this.routeRegexp = this.routeRegexp.replace(RE_ESCAPE, "\\$&");
+    this.routeRegexp = this.routeRegexp.replace(RE_PARAM, (_, mode, name) => {
+      this.paramNames.push(name);
+      return mode === ":" ? "([^/]*)" : "(.*)";
+    });
+    this.routeRegexp = new RegExp("^" + this.routeRegexp + "$");
```
imurchie: String interpolation.

imurchie: Also, until this point, using `this.routeRegexp` seems odd. I'd rather use a local variable to deal with the build up. Something like 
```js
let reStr = route.replace(RE_ESCAPE, "\\$&");
reStr = reStr.replace(RE_PARAM, (_, mode, name) => {
  this.paramNames.push(name);
  return mode === ":" ? "([^/]*)" : "(.*)";
});
this.routeRegexp = new RegExp(`^${reStr}$`);
```

mykola-mokhnach:
@imurchie I've added a couple of unit tests. Btw, what is the command for gulp to execute a single test locally?

imurchie:
The easiest way to only one test is to mark it with `.only`. Otherwise you need to use mocha directly instead of running it through the gulp commands. I don't think there is any provision for using the grep flags through the gulp integration, unfortunately.

lib/mjsonwp/routes.js
```diff
@@ -455,7 +455,55 @@ for (let v of _.values(METHOD_MAP)) {
...
+  }
+}
+
+function routeToCommandName (endpoint, method) {
+  let dstEndpoint = null;
+  for (let currentEndpoint of _.keys(METHOD_MAP)) {
+    const route = new Route(currentEndpoint);
+    // endpoint should already include leading slash
+    // we don't care about the actual session id for matching
+    if (route.parse(`/wd/hub/session/x${endpoint}`) || route.parse(`/wd/hub${endpoint}`) || route.parse(endpoint)) {
```
imurchie: The first string here seems wrong. Shouldn't it be `/wd/hub/session${endpoint}`

imurchie: I think that is why the tests are failing.

mykola-mokhnach: nope - this is by design. Normally the value of _endpoint_ variable is something like "/element" or "/elements" and session id is added automatically. In our case we don't need the real session id, but rather anything just to match the actual route.

imurchie: Then you need another condition that handles needing `/wd/hub/session` appended. Or your test is testing the wrong endpoint value (`'/8b8da8ad-c8c1-42f5-be5a-35ff592beceb/timeouts/implicit_wait'`)

imurchie: Either way, this bit warrants a little more commenting.

mykola-mokhnach: yup, I see now. Will change that

mykola-mokhnach: @imurchie which exact information do you want to see in the comment?

imurchie: Just clarifying what these checks are doing. It might be enough to change `x` to something more informative, like `ignored-session-id`.

mykola-mokhnach:
@imurchie Does the stuff look good now?

imurchie:
Looks good now! Awesome!

imurchie:
Published in `appium-base-driver@2.2.0`.

jlipps:
👍

PR #105, opened 2017-02-17
Add proxy method to verify whether it has a request in progress

mykola-mokhnach:
This is needed to simplify xcuitest-driver logic and get rid of unnecessary flags

### DIFF
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -9,7 +9,7 @@ import { errors } from '../mjsonwp/errors';
 const log = logger.getLogger('JSONWP Proxy');
 // TODO: Make this value configurable as a server side capability
 const LOG_OBJ_LENGTH = 1024; // MAX LENGTH Logged to file / console
-const REQUEST_TIMEOUT = 240000;
+const DEFAULT_REQUEST_TIMEOUT = 240000;
 
 class JWProxy {
   constructor (opts = {}) {
@@ -19,15 +19,32 @@ class JWProxy {
       port: 4444,
       base: '/wd/hub',
       sessionId: null,
-      timeout: REQUEST_TIMEOUT,
+      timeout: DEFAULT_REQUEST_TIMEOUT,
     }, opts);
     this.scheme = this.scheme.toLowerCase();
+    this._activeRequests = [];
   }
 
   // abstract the call behind a member function
   // so that we can mock it in tests
   async request (...args) {
-    return await request(...args);
+    const currentRequest = request(...args);
+    this._activeRequests.push(currentRequest);
+    return await currentRequest.finally(() => _.pull(this._activeRequests, currentRequest));
+  }
+
+  getActiveRequestsCount () {
+    return this._activeRequests.length;
+  }
+
+  cancelActiveRequests () {
+    try {
+      for (let r of this._activeRequests) {
+        r.cancel();
+      }
+    } finally {
+      this._activeRequests = [];
+    }
   }
 
   endpointRequiresSessionId (endpoint) {


lib/jsonwp-proxy/proxy.js
```diff
@@ -19,15 +20,34 @@ class JWProxy {
...
 
   // abstract the call behind a member function
   // so that we can mock it in tests
   async request (...args) {
-    return await request(...args);
+    const currentRequest = request(...args);
+    this._activeRequests.push(r);
+    return await currentRequest.finally(() => {
+                    _.remove(this._activeRequests, (r) => {
+                      r === currentRequest;
```
imurchie: This will always be falsy (and therefore not removed), since it implicitly returns `undefined`. I think you want
```js
_.remove(this._activeRequests, (r) => r === currentRequest);
```

lib/jsonwp-proxy/proxy.js
```diff
@@ -19,15 +20,34 @@ class JWProxy {
...
+                    });
+                  });
+  }
+
+  hasActiveRequests () {
+    return this._activeRequests.length > 0;
```
imurchie: More idiomatically this can be `return !!this._activeRequests.length;`

lib/jsonwp-proxy/proxy.js
```diff
@@ -19,15 +20,34 @@ class JWProxy {
...
     this.scheme = this.scheme.toLowerCase();
+    this._activeRequests = [];
   }
 
   // abstract the call behind a member function
   // so that we can mock it in tests
   async request (...args) {
-    return await request(...args);
+    const currentRequest = request(...args);
+    this._activeRequests.push(r);
```
imurchie: What is `r`?

lib/jsonwp-proxy/proxy.js
```diff
@@ -19,15 +20,34 @@ class JWProxy {
...
+                    });
+                  });
+  }
+
+  hasActiveRequests () {
+    return this._activeRequests.length > 0;
+  }
+
+  async cancelActiveRequests () {
+    await B.Promise.map(this._activeRequests, (r) => {
```
imurchie: Why `map`?

lib/jsonwp-proxy/proxy.js
```diff
@@ -19,15 +20,34 @@ class JWProxy {
...
+      r.cancel();
```
imurchie: Should this be awaited?

lib/jsonwp-proxy/proxy.js
mykola-mokhnach: thanks for the hint

lib/jsonwp-proxy/proxy.js
mykola-mokhnach: Does lint like it? It didn't like this last time I tried

lib/jsonwp-proxy/proxy.js
mykola-mokhnach: to cancel the all active requests at the same time, since cancel() call is synchronous

lib/jsonwp-proxy/proxy.js
imurchie: It might complain about unnecessary `!!`. It would also be perfectly fine to return `this._activeRequests.length` since `0` is falsy and anything else is truthy.

lib/jsonwp-proxy/proxy.js
mykola-mokhnach: [nope](http://bluebirdjs.com/docs/api/cancellation.html)

imurchie: Then a normal iteration is all that is necessary, not a promise one.

mykola-mokhnach: let say we have two active requests in that array. Then calling **cancel** for the first one in a normal loop will block the flow until cancel op is completed. Promise.map is asynchronous and allows to cancel all requests in the array at the same time

lib/jsonwp-proxy/proxy.js
mykola-mokhnach: ok, let it be, but I still like the current approach :-P

lib/jsonwp-proxy/proxy.js
imurchie: That will not work for a synchronous function call, though.

mykola-mokhnach: why not? Everything I want is just to invoke _cancel_ simultaneously for all active requests.

imurchie:
There are linting errors. 
```
[13:36:34] Starting 'eslint'...
[13:36:41] 
/home/travis/build/appium/appium-base-driver/lib/jsonwp-proxy/proxy.js
  35:21  error  Expected indentation of 6 space characters but found 20  indent
  36:19  error  Expected indentation of 4 space characters but found 18  indent
✖ 2 problems (2 errors, 0 warnings)
```

lib/jsonwp-proxy/proxy.js
imurchie: Because with a synchronous call there is no chance for there to be handoff of control. So the first call to the inner method will finish, then the next will be called, then the next, sequentially. There is nothing promise-ey going on in there.

mykola-mokhnach: ok. changed to normal loop.
Thanks for the explanation.

mykola-mokhnach:
Fixed linting errors

imurchie:
Looks good now! Thanks!

imurchie:
Published in `appium-base-driver@2.3.0`.

PR #106, opened 2017-03-03
add event timing framework to base driver

jlipps:
taking a stab at https://github.com/appium/appium/issues/7768

review please @imurchie @dpgraham!

### DIFF
--- a/lib/basedriver/driver.js
+++ b/lib/basedriver/driver.js
@@ -11,6 +11,11 @@ import { util } from 'appium-support';
 
 const NEW_COMMAND_TIMEOUT_MS = 60 * 1000;
 
+const EVENT_SESSION_INIT = 'newSessionRequested';
+const EVENT_SESSION_START = 'newSessionStarted';
+const EVENT_SESSION_QUIT_START = 'quitSessionRequested';
+const EVENT_SESSION_QUIT_DONE = 'quitSessionFinished';
+
 class BaseDriver extends MobileJsonWireProtocol {
 
   constructor (opts = {}, shouldValidateCaps = true) {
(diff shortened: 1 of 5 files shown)


lib/basedriver/driver.js
```diff
@@ -51,6 +56,32 @@ class BaseDriver extends MobileJsonWireProtocol {
...
+   * API method for driver developers to log timings for important events
+   */
+  logEvent (eventName) {
+    if (eventName === "commands") {
+      throw new Error("Cannot log commands directly");
+    }
+    if (typeof eventName !== "string") {
+      throw new Error(`Invalid eventName ${eventName}`);
+    }
+    this._eventHistory[eventName] = Date.now();
```
imurchie: Do we want to only allow a single event of each type? What if we want to have events for WDA startup, which might have to happen multiple times in a session? Would that be the calling code's responsibility, to create reasonable non-clashing event names?

lib/basedriver/driver.js
```diff
@@ -140,6 +171,14 @@ class BaseDriver extends MobileJsonWireProtocol {
   // and ensuring that we execute commands one at a time. This method is called
   // by MJSONWP's express router.
   async executeCommand (cmd, ...args) {
+    // get start time for this command, and log in special cases
+    let startTime = Date.now();
```
imurchie: `process.hrtime([time])` might be a better, if what we care about is intervals. https://blog.tompawlak.org/measure-execution-time-nodejs-javascript

lib/basedriver/driver.js
jlipps: This is a good question. I was thinking we would only want one event of each type, but you raise a good point that an event might reasonably happen more than once and we'd be interested in tracking that. what about having the value of an event key be an array of timestamps rather than a single timestamp?

lib/basedriver/driver.js
jlipps: @imurchie i like process.hrtime, but didn't know of its value when what we are generating at the end of the day is not an interval but an absolute time. i can certainly change it to hrtime, but i think ms resolution is plenty good for the kinds of timings we're thinking about. willing to be persuaded though.

lib/basedriver/driver.js
imurchie: An array of timestamps seems good.

lib/basedriver/driver.js
imurchie: Yeah, probably not necessary. We could, though, just record the elapsed time unless we care about when they started and stopped for some reason.

jlipps: I wasn't sure if all we care about is elapsed time. If not, then we'd have a mix of elapsed times and absolute times in the event log, which seemed odd. OTOH if we just use absolute times everywhere, then a client can figure out what it cares about and determine elapsed time between any two events it cares about. So it seemed more flexible, if a bit more indirect from a client perspective.

Again, I'm willing to be convinced, this just seemed cleanest somehow.

imurchie: I'm convinced!

dpgraham:
Looks good.

PR #108, opened 2017-03-08
Make createSession route compatible to newer Selenium versions

mykola-mokhnach:
See https://github.com/appium/java-client/issues/594 for mode details

### DIFF
--- a/lib/mjsonwp/routes.js
+++ b/lib/mjsonwp/routes.js
@@ -9,7 +9,9 @@ const METHOD_MAP = {
     GET: {command: 'getStatus'}
   },
   '/wd/hub/session': {
-    POST: {command: 'createSession', payloadParams: {required: ['desiredCapabilities'], optional: ['requiredCapabilities', 'capabilities']}}
+    POST: {command: 'createSession',
+           payloadParams: {required: ['desiredCapabilities'],
+                           optional: ['requiredCapabilities', 'capabilities', 'alwaysMatch' , 'firstMatch']}}
   },
   '/wd/hub/sessions': {
     GET: {command: 'getSessions'}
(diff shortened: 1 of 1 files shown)


lib/mjsonwp/routes.js
```diff
@@ -9,7 +9,9 @@ const METHOD_MAP = {
     GET: {command: 'getStatus'}
   },
   '/wd/hub/session': {
-    POST: {command: 'createSession', payloadParams: {required: ['desiredCapabilities'], optional: ['requiredCapabilities', 'capabilities']}}
+    POST: {command: 'createSession',
+           payloadParams: {required: ['desiredCapabilities'],
+                           optional: ['requiredCapabilities', 'capabilities', 'alwaysMatch' , 'firstMatch']}}
```
jlipps: pretty sure that a new client will send in `capabilities` _instead of_ `desiredCapabilities`, and further that `alwaysMatch` and `firstMatch` will be properties of the `capabilities` object, not top-level properties.

not sure if we have a way to specify in base driver that as a requirement we want simply one of a set, rather than all of a set...

mykola-mokhnach: Looking at the [log](https://gist.github.com/TikhomirovSergey/28e7b0ffceacd480f44e1f0127edbfcb), that Mr Sergey kindly provided to us, I could say these properties are on the same level as desiredCapabilites. Anyway, I'd like to try first if this helps as a quick patch, so driver creation does not throw an error. Then we might think about parsing of the actual values.

jlipps:
based on my understanding (https://github.com/appium/java-client/issues/594#issuecomment-285188607) the only change necessary should be to add `capabilities` to the list of optional params.

mykola-mokhnach:
they are already there ;)

mykola-mokhnach:
so I will close this PR, since it's java client, who does not follow the spec :-P

jlipps:
in that case no change should be required. hopefully we can figure out where the issue is in the java client!

PR #109, opened 2017-03-16
Make Mac an accepted platform

penguinho:


jlipps:
this looks good, however it should be merged at the same time as the code which directs appium to use the mac driver if that platform is requested

jlipps:
oh whoops, i didn't realize this was basedriver

jlipps:
published in 2.5.0

PR #113, opened 2017-03-22
record the screen

heeseon:
add two apis
- startRecordingScreen
- stopRecordingScreen


heeseon:
I am not sure why ci failed

test/mjsonwp/routes-specs.js
```diff
@@ -38,39 +38,7 @@ describe('MJSONWP', () => {
       }
       var hash = shasum.digest('hex').substring(0, 8);
       // Modify the hash whenever the protocol has intentionally been modified.
-      hash.should.equal('e2fcfcad');
+      hash.should.equal('be86a90d');
     });
   });
-
```
imurchie: These should not be deleted.

PR #115, opened 2017-03-29
Fix casing on record screen endpoints

imurchie:
Use snake case instead of camel.


imurchie:
Published in `appium-base-driver@2.7.1`.

PR #117, opened 2017-04-06
Implemented capabilities spec

dpgraham:
* Added capabilities.js which implements the capability process spec (https://www.w3.org/TR/webdriver/#dfn-process-capabilities)
* Made 'desiredCapabilities' and 'capabilities' optional parameters, but added new validation logic that checks that at least one of the two was provided


lib/basedriver/capabilities.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  return result;
+}
+
+// Validates caps against a set of constraints
+function validateCaps (caps, constraints = {}, skipPresenceConstraint = false) {
+  let message = '';
+  if (!_.isObject(caps)) {
+    throw new Error(`must be a JSON object`);
```
jlipps: maybe `Caps must be a JSON object`?

lib/basedriver/capabilities.js
```diff
@@ -0,0 +1,111 @@
...
+  if (!_.isArray(allFirstMatchCaps)) {
+    throw new Error('The capabilities.firstMatch argument was not valid for the following reason(s): "capabilities.firstMatch" must be a JSON array or undefined');
+  }
+
+  // Validate the firstMatch capabilities one-by one (see spec #5)
+  let validatedFirstMatchCaps = allFirstMatchCaps.map((firstMatchCaps, index) => {
+    try {
+      return shouldValidateCaps ? validateCaps(firstMatchCaps, constraints) : firstMatchCaps;
+    } catch (e) {
+      throw new Error(`The capabilities.firstMatch argument was not valid for the following reason: "capabilities.firstMatch[${index}] ${e.message}`);
```
jlipps: unclear what the `"` is doing there, maybe it needs a partner?

lib/basedriver/capabilities.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+  });
+
+  // Try to match requiredCaps with one of the validated firstMatch capabilities (see spec #6)
+  let matchedCaps = null;
+  for (let firstMatchCaps of validatedFirstMatchCaps) {
+    try {
+      let matchedCaps = mergeCaps(requiredCaps, firstMatchCaps);
```
jlipps: re-declaring `matchedCaps` here will mean that you're not assigning it to the above `matchedCaps` which it looks like you're trying to do

lib/basedriver/capabilities.js
```diff
@@ -0,0 +1,111 @@
...
+      if (matchedCaps) {
+        break;
+      }
+    } catch (ign) { }
+  }
+
+  // If there was no match, use the de-facto requiredCaps on its own
+  matchedCaps = matchedCaps || requiredCaps;
+  if (!matchedCaps) {
```
jlipps: how is this `if` block any different than the one-liner above?

lib/basedriver/commands/session.js
```diff
@@ -2,16 +2,26 @@ import _ from 'lodash';
 import log from '../logger';
 import { errors } from '../../mjsonwp';
 import UUID from 'uuid-js';
+import { processCapabilities } from '../capabilities';
 
 let commands = {};
 
-commands.createSession = async function (caps) {
+commands.createSession = async function (desiredCapabilities, requiredCaps, capabilities) {
```
jlipps: looks like we're breaking the API here; let's remember to release this as a major rev

lib/mjsonwp/mjsonwp.js
```diff
@@ -66,6 +66,14 @@ function checkParams (paramSets, jsonObj) {
     if (paramSets.optional) {
       optionalParams = paramSets.optional;
     }
+
+    // If a validation function was provided, test it with jsonObj as args
```
jlipps: might be a good idea to tell us a bit more about how `validate` works, i.e.

```js
// If a function was provided as the 'validate' key, it will here be called with 
// jsonObj as the param. If it returns something falsy, verification will be
// considered to have passed. If it returns something else, that will be the
// argument to an error which is thrown to the user

File: lib/basedriver/capabilities.js (hunk 1 above)
Comment (dpgraham) [inline lib/basedriver/capabilities.js]: This error gets caught by other functions and then the error message is concatenated so it will look like `"Invalid argument 'capabilities': must be a JSON object"`

File: lib/basedriver/capabilities.js (hunk 3 above)
Comment (dpgraham) [inline lib/basedriver/capabilities.js]: That's an error, fixing now.

File: lib/basedriver/commands/session.js (hunk 5 above)
Comment (dpgraham) [inline lib/basedriver/commands/session.js]: I think the changes are 'additive'. 
* `caps` is renamed to `desiredCapabilities` to distinguish it from `capabilities`
* `requiredCaps` is unused, it gets passed in by routes.js but doesn't do anything
* `capabilities` is a new parameter

File: lib/basedriver/capabilities.js (hunk 4 above)
Comment (dpgraham) [inline lib/basedriver/capabilities.js]: You're right, that's redundant.

File: lib/basedriver/commands/session.js (hunk 5 above)
Comment (jlipps) [inline lib/basedriver/commands/session.js]: OK cool. So if an existing driver calls `super.createSession` in a current way, nothing will change?

File: lib/basedriver/capabilities.js [hunk 7]
```diff
@@ -0,0 +1,111 @@
+import _ from 'lodash';
+import { validator } from './desired-caps';
+import { util } from 'appium-support';
+
+// Takes primary caps object and merges it into a secondary caps object.
+// (see https://www.w3.org/TR/webdriver/#dfn-merging-capabilities)
+function mergeCaps (primary = {}, secondary = {}) {
+  let result = Object.assign(primary);
```
imurchie: This needs to be `let result = Object.assign({}, primary);` or else it is not doing anything:
```
appium-xcuitest-driver → git master* → node                                                  4.8.2
> a = {s: 't'}
{ s: 't' }
> b = Object.assign(a)
{ s: 't' }
> b
{ s: 't' }
> a
{ s: 't' }
> b.f = 'g'
'g'
> a
{ s: 't', f: 'g' }
> b
{ s: 't', f: 'g' }
> c = Object.assign({}, a)
{ s: 't', f: 'g' }
> c.a = 'b'
'b'
> c
{ s: 't', f: 'g', a: 'b' }
> a
{ s: 't', f: 'g' }
> b
{ s: 't', f: 'g' }
>
```

lib/basedriver/capabilities.js
```diff
@@ -0,0 +1,111 @@
...
+
+  for (let [name, value] of _.toPairs(secondary)) {
+    // Overwriting is not allowed. Primary and secondary must have different properties (w3 rule 4.4)
```
imurchie: w3c

lib/basedriver/capabilities.js
imurchie: This conflation of "object" and JSON object is confusing to me. Might we have a string JSON representation of a JS object? Otherwise it is just an object, not a JSON object, which is nothing, really.

lib/basedriver/capabilities.js
```diff
@@ -0,0 +1,111 @@
...
+  let message = '';
+  if (!_.isObject(caps)) {
+    throw new Error(`must be a JSON object`);
+  }
+
+  constraints = _.cloneDeep(constraints); // Defensive copy
+
+  // Remove the 'presence' constraint if we're not checking for it
+  if (skipPresenceConstraint) {
+    constraints = Object.assign(constraints);
```
imurchie: If this is to make a copy it's already been done on line 28.

lib/basedriver/capabilities.js
```diff
@@ -0,0 +1,111 @@
...
+    for (let key of Object.keys(constraints)) {
+      delete constraints[key].presence;
```
imurchie: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

lib/basedriver/capabilities.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+
+  let validationErrors = validator.validate(_.pickBy(caps, util.hasValue),
+                                              constraints,
+                                              {fullMessages: false});
+                                          
+  if (validationErrors) {
```
imurchie: The `message` variable is localized to this conditional, so declare it here. Also, use an array and then `join` at the end.

lib/basedriver/capabilities.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  requiredCaps = validateCaps(requiredCaps, constraints);
+
+  // Reject 'firstMatch' argument if it's not an array (see spec #3.2)
+  if (!_.isArray(allFirstMatchCaps)) {
+    throw new Error('The capabilities.firstMatch argument was not valid for the following reason(s): "capabilities.firstMatch" must be a JSON array or undefined');
+  }
+
+  // Validate the firstMatch capabilities one-by one (see spec #5)
```
imurchie: one-by-one

lib/basedriver/capabilities.js
imurchie: Does `capabilities.firstMatch[${index}]` have a value? Is it a key? Perhaps it should be logged? This just feels confusing as a log line.

lib/basedriver/capabilities.js
```diff
@@ -0,0 +1,111 @@
...
+
+  // Try to match requiredCaps with one of the validated firstMatch capabilities (see spec #6)
+  let matchedCaps = null;
+  for (let firstMatchCaps of validatedFirstMatchCaps) {
+    try {
+      let matchedCaps = mergeCaps(requiredCaps, firstMatchCaps);
+      if (matchedCaps) {
+        break;
+      }
+    } catch (ign) { }
```
imurchie: If we're ignoring this error here, why even throw it on line 13?

lib/basedriver/capabilities.js
```diff
@@ -0,0 +1,111 @@
...
+    matchedCaps = requiredCaps;
+  }
+
+  // Returns variables for testing purposes
+  return {requiredCaps, allFirstMatchCaps, validatedFirstMatchCaps, matchedCaps};
+}
+
+// Calls parseCaps and just returns the matchedCaps variable
+function processCaps (caps, constraints = {}, shouldValidateCaps = true) {
+  return parseCaps(caps, constraints, shouldValidateCaps).matchedCaps;
```
imurchie: Do we use anything other than `matchedCaps` from this method? If not, why return the rest?

lib/basedriver/capabilities.js
dpgraham: Filter removes values from an array. This doesn't remove values, it deletes a single property from every item.

lib/basedriver/capabilities.js
dpgraham: If it throws an error, the loop continues and looks for a match at the next item. If it doesn't throw an error, we have our first match and we break the loop.

lib/basedriver/capabilities.js
dpgraham: We only used matchedCaps. I probably could refactor it to only return that, the only reason I did so was because I was following the spec and wanted to test it all.

lib/basedriver/capabilities.js
dpgraham: Referring to it as a JSON object is meant to instruct the clients that it accepts a JSON type object.

lib/basedriver/capabilities.js
dpgraham: I'll log out the value instead

lib/basedriver/commands/session.js
dpgraham: That's right. The only problem is that calls to `super.createSession` won't be passing in the 'capabilities' value to the base driver class (unless they did something like `super.createSession.apply(this, [...arguments])`

dpgraham:
Changes pushed

lib/basedriver/capabilities.js
```diff
@@ -0,0 +1,106 @@
...
+  // Validate all of the first match capabilities (see spec #5)
+  let validatedFirstMatchCaps = allFirstMatchCaps.map((firstMatchCaps) => {
+    try {
+      return shouldValidateCaps ? validateCaps(firstMatchCaps, constraints) : firstMatchCaps;
+    } catch (e) {
+      throw new Error(`The capabilities.firstMatch argument was not valid for the following reason: ${firstMatchCaps} ${e.message}`);
+    }
+  });
+
+  // Try to match requiredCaps with first match capabilities, break once it finds it's first match (see spec #6)
```
jlipps: `it's` => `its`

jlipps: you say that this section tries to 'match' caps, but in fact it looks like it's trying to 'merge' caps. is there a difference?

lib/basedriver/capabilities.js
jlipps: since this is only an internal interface, i'm fine returning more vals to make it easy for testing

lib/basedriver/commands/session.js
jlipps: as long as a client could upgrade to a version of base driver containing this change and not have to change anything in order to keep functioning the way it was, then we're good

lib/basedriver/capabilities.js
dpgraham: Merge is a more apt description

PR #118, opened 2017-04-24
Removed adm-zip

dpgraham:
* Added a test for unzipFile (which uses adm-zip)
* Tested that it worked with adm-zip and then failed once adm-zip was taken out
* Replaced 'adm-zip' with zip.extractAllTo from appium-supprt

(reviewers @jlipps @imurchie)

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,8 +23,7 @@
     "lib": "lib"
   },
   "dependencies": {
-    "adm-zip": "^0.4.7",
-    "appium-support": "^2.4.0",
+    "appium-support": "^2.8.0",
     "asyncbox": "^2.3.1",
     "babel-runtime": "=5.8.24",
     "bluebird": "^2.10.2",
@@ -69,6 +68,7 @@
     "eslint-plugin-mocha": "^3.0.0",
     "finalhandler": "^0.4.0",
     "gulp": "^3.9.0",
+    "mock-fs": "^4.2.0",
     "pre-commit": "^1.1.3",
     "serve-static": "^1.10.0",
     "sinon": "^1.17.1"
--- a/test/basedriver/helpers-specs.js
+++ b/test/basedriver/helpers-specs.js
@@ -1,6 +1,10 @@
-import { isPackageOrBundle } from '../../lib/basedriver/helpers';
+import { isPackageOrBundle, unzipFile } from '../../lib/basedriver/helpers';
 import chai from 'chai';
 import chaiAsPromised from 'chai-as-promised';
+import path from 'path';
+import sinon from 'sinon';
+import { system, fs } from 'appium-support';
+import mockFS from 'mock-fs';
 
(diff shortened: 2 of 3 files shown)


test/basedriver/helpers-specs.js
```diff
@@ -15,4 +23,20 @@ describe('helpers', () => {
       isPackageOrBundle('/path/to/an.apk').should.be.false;
     });
   });
+
+  describe('#unzipFile', () => {
+    it('should unzip a .zip file', async () => {
+      const appContents = await fs.readFile(getFixture('FakeIOSApp.app'));
+      await unzipFile(getFixture('FakeIOSApp.app.zip'));
+      await fs.readFile(getFixture('FakeIOSApp.app')).should.eventually.deep.equal(appContents);
```
imurchie: I don't understand what this is asserting. Is seems like it is checking that the `FakeIOSApp.app` didn't change? Is the test just making sure `unzipFile` doesn't error out? If so, it should assert that so it is clear.

test/basedriver/helpers-specs.js
```diff
@@ -15,4 +23,20 @@ describe('helpers', () => {
...
+    });
+
+    it('should unzip a .zip file (force isWindows to be true so we can test the internal zip library)', async () => {
+      const forceWindows = sinon.stub(system, 'isWindows', () => true);
+      const appContents = await fs.readFile(getFixture('FakeIOSApp.app'));
+      await unzipFile(getFixture('FakeIOSApp.app.zip'));
+      await fs.readFile(getFixture('FakeIOSApp.app')).should.eventually.deep.equal(appContents);
```
imurchie: Again.

Maybe I'm misunderstanding how the `unzipFile` function works. Does it overwrite `FakeIOSApp.app`?

dpgraham:
@imurchie I got rid of the test that does not mock `system.isWindows` because that part of the code didn't change and it uses command line operations on the filesystem so it doesn't really work for unit-testing.

I updated the other test to use mock-fs so that it's all done in-memory and it doesn't touch the filesystem. I also changed it so that it hardcodes the base64 zip file into the mock fs, unzips it and then tests the contents.

package.json
```diff
@@ -67,8 +66,10 @@
     "eslint-plugin-babel": "^3.3.0",
     "eslint-plugin-import": "^1.10.2",
     "eslint-plugin-mocha": "^3.0.0",
+    "fancy-log": "^1.3.0",
```
imurchie: I don't see this dep being used.

dpgraham:
Yeah I thought that the dependency was needed to fix an issue with `mock-fs` but it turns out I didn't need it. I tested without it (`rm -rf node_modules && npm install && gulp once`) and it built fine.

dpgraham:
Tagged and published to 2.7.2 (https://www.npmjs.com/package/appium-base-driver)

PR #119, opened 2017-05-06
Add warning message for erroneous responses to make debugging easier

mykola-mokhnach:
This PR has the same reason as https://github.com/appium/appium-xcuitest-driver/pull/432
After further investigation I realised that only this place of code could throw such type of errors.

### DIFF
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -141,8 +141,17 @@ class JWProxy {
         }
       }
     } catch (e) {
+      let responseError;
+      try {
+        responseError = JSON.parse(e.error);
+      } catch (e1) {
+        if (_.isString(e.error) && e.error.length) {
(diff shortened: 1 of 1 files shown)


lib/jsonwp-proxy/proxy.js
```diff
@@ -141,8 +141,17 @@ class JWProxy {
         }
       }
     } catch (e) {
+      let responseError;
+      try {
+        responseError = JSON.parse(e.error);
+      } catch (e1) {
+        if (_.isString(e.error) && e.error.length) {
+          log.warn(`Got unexpected response: ${_.truncate(e.error, 300)}`);
```
imurchie: Truncate length has to be in an object (`_.truncate(e.error, {length: 300})`

mykola-mokhnach: fixed

PR #120, opened 2017-05-18
Add support for optional ‘text’ parameter

mykola-mokhnach:
This parameter is not a part of W3C spec, although Geckodriver has it implemented and thus webdriver.io as well. I assume it won't hurt us if we just have it here instead of failing on _setValue_ call. Check https://github.com/appium/appium/issues/8253#issuecomment-302482837 for more details.

### DIFF
--- a/test/mjsonwp/mjsonwp-specs.js
+++ b/test/mjsonwp/mjsonwp-specs.js
@@ -217,18 +217,32 @@ describe('MJSONWP', async () => {
         json: {id: 'baz', sessionId: 'lol', value: ['a']}
       });
 
-      await request({
-        url: 'http://localhost:8181/wd/hub/session/foo/element/bar/value',
-        method: 'POST',
-        json: {id: 'baz'}
-      }).should.eventually.be.rejectedWith("400");
-
       // make sure adding the optional 'id' doesn't clobber a route where we
       // have an actual required 'id'
       await request({
         url: 'http://localhost:8181/wd/hub/session/foo/frame',
         method: 'POST',
         json: {id: 'baz'}
+      });
+    });
+
+    it('should properly handle variable payload for set value call', async () => {
+      await request({
+        url: 'http://localhost:8181/wd/hub/session/foo/element/bar/value',
+        method: 'POST',
+        json: {sessionId: 'lol'}
+      }).should.eventually.be.rejectedWith("400");
+
+      await request({
+        url: 'http://localhost:8181/wd/hub/session/foo/element/bar/value',
+        method: 'POST',
+        json: {sessionId: 'lol', value: ['a'], text: 'bla'}
+      });
+
+      await request({
+        url: 'http://localhost:8181/wd/hub/session/foo/element/bar/value',
+        method: 'POST',
+        json: {sessionId: 'lol', text: 'bla'}
       });
     });
 
--- a/lib/mjsonwp/routes.js
+++ b/lib/mjsonwp/routes.js
@@ -10,7 +10,7 @@ const METHOD_MAP = {
   },
   '/wd/hub/session': {
     POST: {command: 'createSession', payloadParams: {
-      validate: (jsonObj) => (!jsonObj.capabilities && !jsonObj.desiredCapabilities) && 'we require one of "desiredCapabilities" or "capabilities" object', 
+      validate: (jsonObj) => (!jsonObj.capabilities && !jsonObj.desiredCapabilities) && 'we require one of "desiredCapabilities" or "capabilities" object',
       optional: ['desiredCapabilities', 'requiredCapabilities', 'capabilities']}}
   },
   '/wd/hub/sessions': {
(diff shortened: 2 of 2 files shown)


mykola-mokhnach:
@jlipps how about now?

test/mjsonwp/mjsonwp-specs.js
```diff
@@ -217,18 +217,32 @@ describe('MJSONWP', async () => {
...
 
-      // make sure adding the optional 'id' doesn't clobber a route where we
-      // have an actual required 'id'
       await request({
-        url: 'http://localhost:8181/wd/hub/session/foo/frame',
+        url: 'http://localhost:8181/wd/hub/session/foo/element/bar/value',
         method: 'POST',
-        json: {id: 'baz'}
+        json: {sessionId: 'lol', value: ['a'], text: 'bla'}
+      });
```
jlipps: can we just assert on the response? in this case it should be `a` and not `bla` right? that help me feel that this is doing the right thing.

test/mjsonwp/mjsonwp-specs.js
```diff
@@ -217,18 +217,32 @@ describe('MJSONWP', async () => {
...
+
+      await request({
+        url: 'http://localhost:8181/wd/hub/session/foo/element/bar/value',
+        method: 'POST',
+        json: {sessionId: 'lol', text: 'bla'}
```
jlipps: likewise let's assert that the response is `bla`. (we use a FakeDriver for these tests, which just returns the text value so we can verify it got in correctly)

mykola-mokhnach: the actual response here is now undefined. How do you expect it to look like?

test/mjsonwp/mjsonwp-specs.js
mykola-mokhnach: changed to 

```
      await request({
        url: 'http://localhost:8181/wd/hub/session/foo/element/bar/value',
        method: 'POST',
        json: {sessionId: 'lol', value: ['a'], text: 'bla'}
      }).value.should.eql(['a']);
```

test/mjsonwp/mjsonwp-specs.js
mykola-mokhnach: should I change the FakeDriver implementation?

jlipps: nah. what i'm trying to prove is that no matter whether someone sends in `value` or sends in `text`, that it gets passed to the command appropriately. this should do it now.

jlipps: oh it's undefined? that means something is probably wrong here. the text value should be getting sent to FakeDriver's command appropriately. if that's not happening, then the basedriver code changes here are incorrect.

mykola-mokhnach: doesn't it depend on the particular driver implementation how these payload values are handled? routes protocol just describes the data model format rules

jlipps: yes, and my understanding of this text file is that the particular driver (FakeDriver) is supposed to simply return whatever is passed into it. the fact that it's returning nothing makes me curious whether the value is getting passed into it to begin with (and that _is_ the job of the MJSONWP code)

mykola-mokhnach: This is how we parse route arguments:

```javascript
function makeArgs (reqParams, jsonObj, payloadParams) {
  // we want to pass the url parameters to the commands in reverse order
  // since the command will sometimes want to ignore, say, the sessionId
  let urlParams = _.keys(reqParams).reverse();

  // there can be multiple sets of required params, so find the correct one
  let realRequiredParams = payloadParams.required;
  if (_.isArray(_.first(payloadParams.required))) {
    let keys = _.keys(jsonObj);
    for (let params of payloadParams.required) {
      // check if all the required parameters are in the json object
      if (_.without(params, ...keys).length === 0) {
        // we have all the parameters for this set
        realRequiredParams = params;
        break;
      }
    }
  }
  let args = _.flatten(realRequiredParams).map((p) => jsonObj[p]);
  if (payloadParams.optional) {
    args = args.concat(_.flatten(payloadParams.optional).map((p) => jsonObj[p]));
  }
  args = args.concat(urlParams.map((u) => reqParams[u]));
  return args;
}
```

and this is what is expected by FakeDriver:

```javascript
 async setValue (elementId, value) {
    return value;
  }
```

question - how **text** can be retuned here since if it only expects **value** to be set?

jlipps: this is precisely the right question to ask. because i want to make this change in BaseDriver, but I don't want to make a breaking change. That means that whatever change we make in BaseDriver, it needs to work with the existing function signatures of all the current drivers, including this FakeDriver.

the fact that the parameter is called `value` is inconsequential. basically we need to make sure that regardless of whether the user sends in a `value` or `text` JSON parameter, that it gets into the 2nd positional argument of the command.

in other words, if i send in a JSON request that looks like:

```json
{"value": "foo"}
```

the `args` returned should look like `[<elId>, "foo"]`

and if I send in:

```json
{"text": "foo"}
```

the `args` returned should _still_ look like `[<elId>, "foo"]`

mykola-mokhnach: ok, what do we do with this PR then?

jlipps: fix the code so that the situation I described is true :-) i can dig into this if you're not sure where to begin.

mykola-mokhnach: challenging task, but I'm afraid it will take too long if I pick it up - the code there is quite complicated to extend in the right way, not breaking the backward compatibility %/

jlipps:
going to close in favor of #121

PR #121, opened 2017-05-26
allow both 'value' and 'text' params for setValue

jlipps:
this introduces a 'makeArgs' option for the spec which allows handling of more complex scenarios.

@mykola-mokhnach this should do the trick.

### DIFF
--- a/lib/mjsonwp/mjsonwp.js
+++ b/lib/mjsonwp/mjsonwp.js
@@ -106,39 +106,60 @@ function checkParams (paramSets, jsonObj) {
   throw new errors.BadParametersError(paramSets, receivedParams);
 }
 
-function makeArgs (reqParams, jsonObj, payloadParams) {
-  // we want to pass the url parameters to the commands in reverse order
-  // since the command will sometimes want to ignore, say, the sessionId
-  let urlParams = _.keys(reqParams).reverse();
+/*
+ * This method takes 3 pieces of data: request parameters ('requestParams'),
+ * a request JSON body ('jsonObj'), and 'payloadParams', which is the section
+ * from the route definition for a particular endpoint which has instructions
+ * on handling parameters. This method returns an array of arguments which will
+ * be applied to a command.
+ */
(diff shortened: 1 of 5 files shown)


lib/mjsonwp/mjsonwp.js
```diff
@@ -124,10 +124,20 @@ function makeArgs (reqParams, jsonObj, payloadParams) {
...
     }
   }
-  let args = _.flatten(realRequiredParams).map((p) => jsonObj[p]);
-  if (payloadParams.optional) {
-    args = args.concat(_.flatten(payloadParams.optional).map((p) => jsonObj[p]));
+  let args;
+  if (_.isFunction(payloadParams.makeArgs)) {
+    // if the spec defines a 'makeArgs' function, allow it to figure out how to
+    // turn the json inputs into a command arguments array
+    args = payloadParams.makeArgs(jsonObj);
```
mykola-mokhnach: I would expect more detailed description of makeArgs property here - what it should do and what is the expected result/the type of it

lib/mjsonwp/routes.js
```diff
@@ -135,7 +135,21 @@ const METHOD_MAP = {
     GET: {command: 'getText'}
   },
   '/wd/hub/session/:sessionId/element/:elementId/value': {
-    POST: {command: 'setValue', payloadParams: {required: ['value']}}
+    POST: {command: 'setValue', payloadParams: {
+      validate: (jsonObj) => {
```
mykola-mokhnach: you don't like "short" lambda form like 

```
(jsonObj) => (!jsonObj.value && !jsonObj.text) &&  'we require one of "text" or "value" params'
```

do you?

jlipps: I do when it can stay within 80 chars length or so. it was the line length that made me want to chop it

lib/mjsonwp/mjsonwp.js
jlipps: ok, i will adjust the comment

PR #123, opened 2017-06-14
Add route to get current package

imurchie:
`appium` vendor route for getting the current package, mirroring the getting of the current activity, which is already available.

First step in https://github.com/appium/appium/issues/5493


imurchie:
Published in `appium-base-driver@2.10.0`.

PR #124, opened 2017-06-15
fix bug in DeviceSettings where we could mutate external object

jlipps:


jlipps:
published in `appium-base-driver@2.10.1`

PR #125, opened 2017-07-06
Allow downloaded app urls to have query strings

imurchie:
When downloading an app, use the URL's pathname rather than the whole URL to determine if it needs to be unzipped.

Fixes https://github.com/appium/appium/issues/8793

### DIFF
--- a/lib/basedriver/helpers.js
+++ b/lib/basedriver/helpers.js
@@ -18,7 +18,7 @@ async function configureApp (app, appExt, mountRoot="Volumes", windowsShareUserN
   }
 
   let newApp = null;
-  let shouldUnzipApp = _.includes(ZIP_EXTS, path.extname(app));
+  let shouldUnzipApp = false;
 
   // check if we're copying from a windows network share
   if (_.startsWith(app, "\\")) {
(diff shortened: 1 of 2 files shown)


lib/basedriver/helpers.js
```diff
@@ -27,19 +27,14 @@ async function configureApp (app, appExt, mountRoot="Volumes", windowsShareUserN
 
   if ((app || '').substring(0, 4).toLowerCase() === 'http') {
     logger.info(`Using downloadable app '${app}'`);
-    let downloadedApp = await downloadApp(app, shouldUnzipApp ? '.zip' : appExt);
+    let downloadedApp = await downloadApp(app, appExt);
     newApp = downloadedApp.targetPath;
```
mykola-mokhnach: wouldn't deconstruction more readable here?

imurchie: Would it? There would be a name mismatch (since the semantics of the returned path and what the calling code uses it for are distinct) so there would still be an assignment outside the deconstruction step. Since it's a mix of local variables and ones from the enclosing scope it would also need another step:
```js
let targetPath;
({targetPath, shouldUnzipApp} = await downloadApp(app, appExt));
newApp = targetPath;
```

I do think, semantically, the `shouldUnzipApp` of the returned object needs to be renamed, so part of that would be gone. Something like:
```js
let {targetPath, isZipFile} = await downloadApp(app, appExt);
newApp = targetPath;
shouldUnzipApp = isZipFile;
```

mykola-mokhnach:
This might be a bit tricky, since APK and ipa files are internally zip archives as well. I'm just curios about whether one can apply some universal behaviour based on downloaded file magic

imurchie:
We already handle that logic. This is just stripping away the query string when applying it.

lib/basedriver/helpers.js
mykola-mokhnach: maybe I'm wrong, but how about this: http://wesbos.com/destructuring-renaming/

imurchie: Ah, yes. The directionality of that is backwards for my mind so I always ignore it. Alas.

imurchie:
Published in `appium-base-driver@2.10.2`.

PR #126, opened 2017-07-18
exchange deleteSession and curCommandCancellable

cool-firer:
session should be closed firstly if some unexpected happens, so the following request can be responsed immediately.


cool-firer:
if something caused unexpected shutdown while appium is handling one request, the request is canceled and start to run deleteSession. but if second request comes at the same time, appium may handle request before deleteSession is done. So it's better to put deleteSesion at first.

PR #127, opened 2017-07-18
warn when a setting does not exist

jlipps:
simple fix for https://github.com/appium/appium/issues/3585


imurchie:
Published in `appium-base-driver@2.11.0`.

PR #128, opened 2017-08-01
Make different pages and links distinguishable

imurchie:
We have tests that rely on distinguishing between what links we tap, but most of the links actually open the same page. So change that, so each link opens a distinct page.

This will be a breaking change, unfortunately, since tests that rely on the old behavior will fail unexpectedly with the text they are searching for having been changed.


imurchie:
I guess we could, yes!

imurchie:
Moved to added links instead of changed. Tested against ios and xcuitest drivers and nothing seems broken.

PR #129, opened 2017-08-07
Add new endpoints

vrunoa:
Adding endpoints for the new methods we're adding on android emulators.

### DIFF
--- a/lib/mjsonwp/routes.js
+++ b/lib/mjsonwp/routes.js
@@ -10,7 +10,7 @@ const METHOD_MAP = {
   },
   '/wd/hub/session': {
     POST: {command: 'createSession', payloadParams: {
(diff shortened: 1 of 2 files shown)


vrunoa:
thanks @imurchie. 
Once the emu methods are merge on appium-android-driver we can merge this one.

imurchie:
The order should not matter. Any method here that is not implemented on the driver will just return a "not yet implemented" error automatically.

vrunoa:
Oh didnt knew that. Cool.

imurchie:
That's how we handle things like the fact that the iOS drivers are not going to implement these methods.

Just for interest's sake, the code is here: https://github.com/appium/appium-base-driver/blob/master/lib/basedriver/driver.js#L182-L184

imurchie:
Published in `appium-base-driver@2.12.0`.

PR #133, opened 2017-08-15
start_recording_screen, stop_recording_screen

heeseon:
to startRecordingScreen, stopRecordingScreen


imurchie:
Routes use snake case, not camel case. This change will not be accepted.

PR #134, opened 2017-08-17
Add a possibility to execute commands without queueing them

mykola-mokhnach:
This is a "blind" implementation of the feature that we discussed before. I didn't test it with a real client, but it's just how I see it should look like. The next step will be to override _isCommandsQueueEnabled_ property in AppiumDriver to make it return false :)

### DIFF
--- a/lib/basedriver/driver.js
+++ b/lib/basedriver/driver.js
@@ -66,7 +66,7 @@ class BaseDriver extends MobileJsonWireProtocol {
 
   /**
    * This property is used by AppiumDriver to store the data of the
-   * specific driver sessions. This datacan be later used to adjust
+   * specific driver sessions. This data can be later used to adjust
    * properties for driver instances running in parallel.
    * Override it in inherited driver classes if necessary.
    *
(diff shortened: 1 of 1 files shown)


lib/basedriver/driver.js
```diff
@@ -195,33 +210,41 @@ class BaseDriver extends MobileJsonWireProtocol {
...
-      // have an unexpected shutdown event, for example, we can cancel it from
-      // outside, rejecting the current command immediately
-      this.curCommandCancellable = new B((r) => { r(); }).then(() => { // eslint-disable-line promise/prefer-await-to-then
-        return this[cmd](...args);
-      }).cancellable();
-      return this.curCommandCancellable;
-    });
-    this.curCommand = nextCommand.catch(() => {});
-    let res = await nextCommand;
+      res = await B.resolve(() => this[cmd](...args));
```
jlipps: why not just:

```js
res = await this[cmd](...args);
```
?

mykola-mokhnach: if you say _await_ is smart enough to process this then B.resolve can be removed

jlipps:
Looks great now. I think we should find someone to do some real-world testing of this branch before merging.

mykola-mokhnach:
It's safe to merge it, since the default behaviour is preserved. The only thing, which might be dangerous to merge, is override of the property in AppiumDriver

jlipps:
published in `appium-base-driver@2.15.0`

PR #136, opened 2017-08-25
Detect a non-working chromedriver connection after androidDriver.closeApp() and androidDriver.launchApp()

cschabetsberger-micro-focus:
Whenever I executed androidDriver.closeApp() and androidDriver.launchApp() for my hybrid app, I had the situation that Appium did not detect that a new chrome driver connection has to be established. I received then the error message "chrome not reachable". This can be fixed by checking the response body status code: resBody should only be returned as a whole, when there was no resonse body status code defined in the response.

### DIFF
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -166,7 +166,7 @@ class JWProxy {
     if (_.includes(statusCodesWithRes, response.statusCode)) {
       if (response.statusCode === 200 && resBody.status === 0) {
         return resBody.value;
-      } else if (response.statusCode === 200) {
+      } else if (response.statusCode === 200 && _.isUndefined(resBody.status)) {
         return resBody;
       }
       let message = getSummaryByCode(resBody.status);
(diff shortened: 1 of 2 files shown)


imurchie:
I don't understand how this does anything but cover up the problem? Shouldn't there be fix in `appium-android-driver`?

cschabetsberger-micro-focus:
Sorry for the delay, for some reason I missed that you already updated this request.

I made the fix there because the communication response is handled there and from the code it seemed to me that if the statusCode is 200, only the resBody.status == 0 is considered as successful (or if there is no resBody.status defined). In my situation, chromeDriver reported a resBody.status != 0 (I don't remember the exact status code) and it reported an error message. So I think an error should be thrown there if resBody.status != 0.

I did a lot of testing in the meantime and this seems to fix all stability issues I had with closeApp and launchApp.

imurchie:
Please run the unit tests locally (`npm run test`). There is a broken test from this change.

cschabetsberger-micro-focus:
Sorry, I will take a look at the failing test and file another PR.

imurchie:
No other PR is necessary. Just push another commit to your `bugfix-chromedriver-connection-issue` branch and it will update here.

imurchie:
Thanks!

imurchie:
Published in `appium-base-driver@2.15.7`.

PR #138, opened 2017-09-07
Add static page with app banner

imurchie:
So we can test what happens when a page has the app download banner at the top.


imurchie:
Published in `appium-base-driver@2.15.3`

PR #139, opened 2017-09-21
fix the smbfs path

wei-gu:


imurchie:
@wei-gu Can you check your git configuration? You don't seem to be committing with a username associated with a github account.

imurchie:
Bringing @yaphatak into the conversation, since he wrote the original code and might know more.

wei-gu:
@yaphatak left Microsoft and I took the work from him. For now, we are using workaround to make it work with: 
"**windowsShareUserName**":"**//**domain;name"
to add extra "**//**"

jlipps:
ok, sounds good. looks like all that's left is fixing your git config and re-pushing

wei-gu:
just did and create another PR

PR #142, opened 2017-10-02
Add session id to base driver log records

mykola-mokhnach:
This Pr is a part of https://github.com/appium/appium/issues/9080

### DIFF
--- a/lib/basedriver/commands/session.js
+++ b/lib/basedriver/commands/session.js
@@ -1,5 +1,4 @@
 import _ from 'lodash';
-import log from '../logger';
 import { errors } from '../../mjsonwp';
 import UUID from 'uuid-js';
 import { processCapabilities } from '../capabilities';
@@ -8,8 +7,8 @@ let commands = {};
 
 commands.createSession = async function (desiredCapabilities, requiredCaps, capabilities) {
   if (this.sessionId !== null) {
-    throw new errors.SessionNotCreatedError('Cannot create a new session ' +
-                                            'while one is in progress');
+    this.log.errorAndThrow(new errors.SessionNotCreatedError('Cannot create a new session ' +
+                                                             'while one is in progress'));
   }
 
   let caps;
@@ -19,7 +18,7 @@ commands.createSession = async function (desiredCapabilities, requiredCaps, capa
     caps = processCapabilities(capabilities, this.desiredCapConstraints, this.shouldValidateCaps);
   }
 
-  caps = fixCaps(caps, this.desiredCapConstraints);
+  caps = fixCaps(caps, this.desiredCapConstraints, this.log);
   this.validateDesiredCaps(caps);
 
   this.sessionId = UUID.create().hex;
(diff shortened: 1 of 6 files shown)


lib/basedriver/driver.js
```diff
@@ -62,6 +61,10 @@ class BaseDriver extends MobileJsonWireProtocol {
     this._eventHistory = {
       commands: [] // commands get a special place
     };
+
+    this.log = logger.getLogger(
+      () => `${this.constructor.name}` +
+            (_.isString(this.sessionId) && this.sessionId.length >= 8) ? ` [${this.sessionId.substring(0, 8)}]` : '');
```
imurchie: Parentheses are mismatched in here. This is breaking the unit tests, too. (Why are the unit tests not run locally? This would be found.)

mykola-mokhnach: Fixed

mykola-mokhnach:
@imurchie Please also let me know where this instance logger can also be added instead of the static one

imurchie:
I don't know what you mean.

_No_ driver uses this, and some drivers use multiple loggers in order to get distinct prefixes. So this, as currently envisioned, is not going to work.

mykola-mokhnach:
I mean from which extensions/commands we should also replace the static logger with the instance one.

lib/basedriver/commands/session.js
```diff
@@ -89,7 +88,7 @@ commands.deleteSession = async function (/* sessionId */) {
   this.sessionId = null;
 };
 
-function fixCaps (originalCaps, desiredCapConstraints = {}) {
+function fixCaps (originalCaps, desiredCapConstraints = {}, logger = {}) {
```
imurchie: Having this default on `logger` will cause it to fail when used (since `{}` doesn't have a `warn` function on it) if there is no logger passed in. You'd need something like `, logger = {warn () {}}`

imurchie:
I don't know which ones. I would presume _every_ log line should be changed. That seems like a big project, and would need some thinking, since this solution won't work. I think it would be better to have a function to get a logger with a particular prefix and session id on it, which can be called to populate the logger on the driver. I guess then on sub-loggers there would be a need to pass the session id around.

jlipps:
Yeah, something like `this.log.info(...)` from a driver perspective, rather than just `log.info`, to scope the logs to the session. there's also the problem of other modules that don't even belong to the driver itself, like for example logging from appium-ios-simulator. it doesn't know anything about sessions. so it would have to be parametrized with a log prefix, or a logger passed in, or something. there are probably economical ways to do this but as isaac said it's a lot of work.

lib/basedriver/driver.js
```diff
@@ -62,6 +61,17 @@ class BaseDriver extends MobileJsonWireProtocol {
     this._eventHistory = {
       commands: [] // commands get a special place
     };
+
+    this.initLogger();
+  }
+
+  initLogger () {
+    const instanceId = Math.random().toString(16).substring(2, 8);
```
jlipps: sessionId is already a UUID, why do we care about an instance ID in addition?

mykola-mokhnach: because session id is only created after we have a session and instance id exists when driver instance is constructed/session is deleted.

mykola-mokhnach:
My initial intension was to implement it as MVP without serious changes to the existing codebase. I don't see any problems if we close this PR, so you could implement it better.

jlipps:
I'm definitely open to merging this if we think it is a good base for future work. I was just agreeing with @imurchie that by itself this doesn't provide a lot of value since the drivers themselves aren't automatically getting the advantage of the work you've done in this PR.

PR #144, opened 2017-11-01
Adding capabilities for visual testing of native apps.

danielputerman:
This PR is part of a number of PRs for Appium (additional ones are for uiautomator2-driver/server, and ios-driver), adding capabilities which allow for visual testing of native apps. These capabilities include:
1. Viewport related capabilities: status bar height, viewport rect (position/size), getting viewport screenshot.
3. Additional functionality for scrollable elements: finding scrollable views in the application, retrieving their content size and providing deterministic scrolling inside them.

Specifically this PR includes image cropping functionality for getting viewport images, used for both Android and iOS in the image-helpers.js. Originally, it included updated routes, but most of the new commands were added under "mobile: ", so no actual routes update was necessary.


lib/basedriver/image-helpers.js
```diff
@@ -0,0 +1,32 @@
+import {PNG} from 'pngjs';
+
+function cropBase64Image (base64EncodedImage, rect) {
+  let pngImage = PNG.sync.read(Buffer.from(base64EncodedImage, 'base64'));
+
+  if (rect.top < 0 || rect.top >= pngImage.height || rect.left < 0 || rect.left >= pngImage.width) {
```
mykola-mokhnach: I'd rather implement this logic in the following way (pseudo-cde):
```
const imageRect = ...
const intersectionRect = getRectIntersection(rect, imageRect)
if (isEmptyRect(getRectIntersection())) {
   throw error
}
return cropImage(image, intersectionRect);
```

lib/basedriver/image-helpers.js
```diff
@@ -0,0 +1,32 @@
...
+    throw new Error('rect is outside the image bounds!');
+  }
+
+  const yStart = rect.top,
+        yEnd = Math.min(rect.top + rect.height, pngImage.height),
+        xStart = rect.left,
+        xEnd = Math.min(rect.left + rect.width, pngImage.width);
+
+  let y, x, idx, i, croppedArray = [];
+  for (y = yStart; y < yEnd; y++) {
```
imurchie: We can scope some of these variables in the `for` loops:
```js
let croppedArray = [];
for (let y = yStart; y < yEnd; y++) {
  for (let x = xStart; x < xEnd; x++) {
    let idx = (pngImage.width * y + x) << 2;
    for (let i = 0; i < 4; i++) {
      croppedArray.push(pngImage.data[idx + i]);
    }
  }
}
```

lib/basedriver/image-helpers.js
```diff
@@ -0,0 +1,32 @@
+import {PNG} from 'pngjs';
+
+function cropBase64Image (base64EncodedImage, rect) {
+  let pngImage = PNG.sync.read(Buffer.from(base64EncodedImage, 'base64'));
```
imurchie: Is there no asynchronous method for doing this?

test/basedriver/image-helpers-specs.js
```diff
@@ -0,0 +1,43 @@
+/* eslint-disable */
+import { cropBase64Image } from '../../lib/basedriver/image-helpers';
+import {PNG} from 'pngjs';
```
imurchie: `{ PNG }`

astappev:
Updated the helper according to suggestions

lib/basedriver/image-helpers.js
```diff
@@ -1,32 +1,69 @@
-import {PNG} from 'pngjs';
+import { Buffer } from 'buffer';
+import { PNG } from 'pngjs';
+import B from 'bluebird';
 
-function cropBase64Image (base64EncodedImage, rect) {
-  let pngImage = PNG.sync.read(Buffer.from(base64EncodedImage, 'base64'));
+async function createImageFromBase64 (base64EncodedImage) {
+  const buffer = Buffer.from(base64EncodedImage, 'base64');
```
mykola-mokhnach: this method is only available since node 8

lib/basedriver/image-helpers.js
```diff
@@ -1,32 +1,69 @@
...
+function getRectIntersection (rect, imageSize) {
+  const left = rect.left >= imageSize.width ? imageSize.width : rect.left;
+  const top = rect.top >= imageSize.height ? imageSize.height : rect.top;
+  const width = imageSize.width >= (left + rect.width) ? rect.width : 0;
+  const height = imageSize.height >= (top + rect.height) ? rect.height : 0;
+  return { left, top, width, height };
+}
 
-  return PNG.sync.write(pngImage).toString('base64');
+function isValidRect (rect) {
```
mykola-mokhnach: isEmptyRect might be more readable

lib/basedriver/image-helpers.js
```diff
@@ -1,32 +1,69 @@
...
-      idx = (pngImage.width * y + x) << 2;
-      for (i = 0; i < 4; i++) {
-        croppedArray.push(pngImage.data[idx + i]);
+  const yEnd = Math.min(rect.top + rect.height, image.height);
+  const xEnd = Math.min(rect.left + rect.width, image.width);
+
+  let croppedArray = [];
+  for (let y = rect.top; y < yEnd; y++) {
+    for (let x = rect.left; x < xEnd; x++) {
+      const idx = (image.width * y + x) << 2;
```
mykola-mokhnach: is it ok that we use _image.width_ to generate the data and then change it?

lib/basedriver/image-helpers.js
```diff
@@ -1,32 +1,69 @@
...
+      reject(err);
+    });
+  });
+}
 
-  if (rect.top < 0 || rect.top >= pngImage.height || rect.left < 0 || rect.left >= pngImage.width) {
+function cropImage (image, rect) {
+  const imageRect = {width: image.width, height: image.height};
+  const intersectionRect = getRectIntersection(rect, imageRect);
+  if (!isValidRect(intersectionRect)) {
```
mykola-mokhnach: is it expected that intersectionRect variable is only used to check if the intersection is valid and is not used for further calculations at all?

astappev: No, it is mistake

lib/basedriver/image-helpers.js
astappev: Changed to default buffer constructor

lib/basedriver/image-helpers.js
astappev: Ok, renamed

lib/basedriver/image-helpers.js
astappev: It is ok because we override it only after the loop, but refactored to be more clear

lib/basedriver/image-helpers.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  const yEnd = Math.min(interRect.top + interRect.height, imageRect.height);
+  const xEnd = Math.min(interRect.left + interRect.width, imageRect.width);
+
+  let croppedArray = [];
+  for (let y = interRect.top; y < yEnd; y++) {
+    for (let x = interRect.left; x < xEnd; x++) {
+      const idx = (imageRect.width * y + x) << 2;
+      for (let i = 0; i < 4; i++) {
```
mykola-mokhnach: I don't really like these magic numbers. Can we have a set of constants instead whose names explain the purpose of each number?

lib/basedriver/image-helpers.js
```diff
@@ -0,0 +1,75 @@
+import { Buffer } from 'buffer';
+import { PNG } from 'pngjs';
+import B from 'bluebird';
+
+async function cropBase64Image (base64EncodedImage, rect) {
```
mykola-mokhnach: It would be useful to add doctrings on these all helper methods Check https://github.com/appium/appium-adb/blob/9ee5b9251fb15cc2c0ae3e86b6d1659d6aa576cf/lib/tools/adb-commands.js on docstring examples

lib/basedriver/image-helpers.js
```diff
@@ -0,0 +1,75 @@
...
+  const imageRect = {width: image.width, height: image.height};
+  const interRect = getRectIntersection(rect, imageRect);
+  if (isEmptyRect(interRect)) {
+    throw new Error('rect is outside the image bounds!');
+  }
+
+  const yEnd = Math.min(interRect.top + interRect.height, imageRect.height);
+  const xEnd = Math.min(interRect.left + interRect.width, imageRect.width);
+
+  let croppedArray = [];
```
mykola-mokhnach: also const

lib/basedriver/image-helpers.js
```diff
@@ -0,0 +1,75 @@
...
+  return new B((resolve, reject) => {
+    const image = new PNG({filterType: 4});
+    image.parse(imageBuffer, (err, image) => { // eslint-disable-line promise/prefer-await-to-callbacks
+      if (err) { return reject(err); }
+      resolve(image);
+    });
+  });
+}
+
+async function createBase64FromImage (image) {
```
mykola-mokhnach: imageToBase64Buffer

lib/basedriver/image-helpers.js
```diff
@@ -0,0 +1,75 @@
...
+  const croppedImage = cropImage(image, rect);
+  return await createBase64FromImage(croppedImage);
+}
+
+async function createImageFromBase64 (base64EncodedImage) {
+  const buffer = new Buffer(base64EncodedImage, 'base64');
+  return await createImageFromBuffer(buffer);
+}
+
+async function createImageFromBuffer (imageBuffer) {
```
mykola-mokhnach: bufferToImage

lib/basedriver/image-helpers.js
```diff
@@ -0,0 +1,75 @@
...
+import { PNG } from 'pngjs';
+import B from 'bluebird';
+
+async function cropBase64Image (base64EncodedImage, rect) {
+  const image = await createImageFromBase64(base64EncodedImage);
+  const croppedImage = cropImage(image, rect);
+  return await createBase64FromImage(croppedImage);
+}
+
+async function createImageFromBase64 (base64EncodedImage) {
```
mykola-mokhnach: base64BufferToImage

astappev:
Updated

lib/basedriver/image-helpers.js
```diff
@@ -2,18 +2,27 @@ import { Buffer } from 'buffer';
 import { PNG } from 'pngjs';
 import B from 'bluebird';
 
-async function cropBase64Image (base64EncodedImage, rect) {
-  const image = await createImageFromBase64(base64EncodedImage);
+/**
+ * Crop the image by given rectangle (use base64 string as input and output)
+ *
+ * @param {string} base64Image The string with base64 encoded image
+ * @param {{left: number, top: number, width: number, height: number}} rect The selected region of image
```
mykola-mokhnach: use _typedef_ annotation to describe complex structures. Check https://github.com/appium/appium-adb/blob/a3b5afcd44f4670eaa745faea1ea478b7134b323/lib/tools/adb-commands.js#L774 for example

lib/basedriver/image-helpers.js
```diff
@@ -2,18 +2,27 @@ import { Buffer } from 'buffer';
...
+ * @return {string} base64 encoded string of cropped image
```
mykola-mokhnach: string or buffer?

lib/basedriver/image-helpers.js
```diff
@@ -23,7 +32,13 @@ async function createImageFromBuffer (imageBuffer) {
   });
 }
 
-async function createBase64FromImage (image) {
+/**
+ * Create a base64 string for given image object
+ *
+ * @param {PNG} image The image object
+ * @return {string} The string with base64 encoded image
```
mykola-mokhnach: buffer

lib/basedriver/image-helpers.js
```diff
@@ -34,29 +49,39 @@ async function createBase64FromImage (image) {
...
-  for (let y = interRect.top; y < yEnd; y++) {
-    for (let x = interRect.left; x < xEnd; x++) {
-      const idx = (imageRect.width * y + x) << 2;
-      for (let i = 0; i < 4; i++) {
-        croppedArray.push(image.data[idx + i]);
+  const croppedArray = [];
+  for (let y = firstVerticalPixel; y < lastVerticalPixel; y++) { // loop over vertical pixels
+    for (let x = firstHorizontalPixel; x < lastHorizontalPixel; x++) { // loop over horizontal pixels
+      const idx = (imageRect.width * y + x) << 2; // first byte location of current pixel's block
+      for (let i = 0; i < 4; i++) { // each pixel's block consists of 4 bytes
```
mykola-mokhnach: proper variable names will make these comments unnecessary, for example

`for (let i = 0; i < 4; i++)` -> `for (let byteIdx = 0; byteIdx < BYTES_IN_BLOCK; ++byteIdx)`

lib/basedriver/image-helpers.js
```diff
@@ -34,29 +49,39 @@ async function createBase64FromImage (image) {
...
-  let croppedArray = [];
-  for (let y = interRect.top; y < yEnd; y++) {
-    for (let x = interRect.left; x < xEnd; x++) {
-      const idx = (imageRect.width * y + x) << 2;
-      for (let i = 0; i < 4; i++) {
-        croppedArray.push(image.data[idx + i]);
+  const croppedArray = [];
+  for (let y = firstVerticalPixel; y < lastVerticalPixel; y++) { // loop over vertical pixels
+    for (let x = firstHorizontalPixel; x < lastHorizontalPixel; x++) { // loop over horizontal pixels
+      const idx = (imageRect.width * y + x) << 2; // first byte location of current pixel's block
```
mykola-mokhnach: idx -> firstByteIdxInPixelBlock

lib/basedriver/image-helpers.js
astappev: A driver's getScreenshot command returns a string, therefore it should be a string.

lib/basedriver/image-helpers.js
astappev: See comment above

lib/basedriver/image-helpers.js
```diff
@@ -0,0 +1,110 @@
+import { Buffer } from 'buffer';
+import { PNG } from 'pngjs';
+import B from 'bluebird';
+
+/**
+ * @typedef {Object} Region
```
mykola-mokhnach: 👍

lib/basedriver/image-helpers.js
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * @param {PNG} image The image to crop
+ * @param {Region} rect The selected region of image
+ * @return {PNG} Cropped image
+ */
+function cropImage (image, rect) {
+  const imageRect = {width: image.width, height: image.height};
+  const interRect = getRectIntersection(rect, imageRect);
+  if (isEmptyRect(interRect)) {
+    throw new Error('rect is outside the image bounds!');
```
mykola-mokhnach:
```
`Cannot crop ${JSON.stringify(rect)} from ${JSON.stringify(imageRect)}, because the intersection between them is empty`
```

lib/basedriver/image-helpers.js
```diff
@@ -0,0 +1,110 @@
...
+  return await imageToBase64(croppedImage);
+}
+
+/**
+ * Create a pngjs image from given base64 image
+ *
+ * @param {string} base64Image The string with base64 encoded image
+ * @return {PNG} The image object
+ */
+async function base64ToImage (base64Image) {
```
mykola-mokhnach: Please do not forget to update these method names in PRs for dependent drivers

astappev:
I merged all my commits into one and found that no other changes left in the pull request. All other commits add some staff and remove it later.

I created new [pull request](https://github.com/appium/appium-base-driver/pull/148) to be more clear

lib/basedriver/image-helpers.js
```diff
@@ -0,0 +1,110 @@
...
+  const firstVerticalPixel = interRect.top;
+  const lastVerticalPixel = interRect.top + interRect.height;
+
+  const firstHorizontalPixel = interRect.left;
+  const lastHorizontalPixel = interRect.left + interRect.width;
+
+  const croppedArray = [];
+  for (let y = firstVerticalPixel; y < lastVerticalPixel; y++) {
+    for (let x = firstHorizontalPixel; x < lastHorizontalPixel; x++) {
+      const firstByteIdxInPixelBlock = (imageRect.width * y + x) << 2;
```
jlipps: is there any reason to use the bitshift operator instead of simply multiplying by `BYTES_IN_PIXEL_BLOCK` which appears to do the same thing? maybe i don't understand image math :-)

lib/basedriver/image-helpers.js
```diff
@@ -0,0 +1,110 @@
...
+  image.height = interRect.height;
+  return image;
+}
+
+function getRectIntersection (rect, imageSize) {
+  const left = rect.left >= imageSize.width ? imageSize.width : rect.left;
+  const top = rect.top >= imageSize.height ? imageSize.height : rect.top;
+  const width = imageSize.width >= (left + rect.width) ? rect.width : 0;
+  const height = imageSize.height >= (top + rect.height) ? rect.height : 0;
+  return { left, top, width, height };
```
jlipps: `return {left, top, width, height}` (we only put spaces around braces for imports, though that may have been a bad choice!)

lib/basedriver/image-helpers.js
```diff
@@ -0,0 +1,110 @@
...
+  image.data = new Buffer(croppedArray);
+  image.width = interRect.width;
+  image.height = interRect.height;
+  return image;
+}
+
+function getRectIntersection (rect, imageSize) {
+  const left = rect.left >= imageSize.width ? imageSize.width : rect.left;
+  const top = rect.top >= imageSize.height ? imageSize.height : rect.top;
+  const width = imageSize.width >= (left + rect.width) ? rect.width : 0;
```
jlipps: it looks like what this is doing is setting width to 0 even if there is a partial intersection. we want to disallow partial intersections? (rather than say resizing the crop rect to just go to the edge of the image?)

i don't care either way, just want to make sure we know what is going on. some comments might be useful here. or help me understand if i'm getting the logic wrong.

lib/basedriver/image-helpers.js
astappev: Yes, it should work in same way. But, I think bitshift should be faster, maybe I'm wrong, it's javascript :)

lib/basedriver/image-helpers.js
astappev: Right now, the method accepts only regions that fully satisfied by given image. Otherwise, we set width (or height) to 0 and then throw an error (so, we use 0 as flag that something wrong with the region).

For the methods in which we use the method we don't need to support partial intersections, bacause we always know a size of image. If you think whould be better to support partial intersections, I can update the code.

jlipps: nope, no need! just making sure we were OK with this.

jlipps:
closing since a new PR came in

PR #145, opened 2017-11-03
W3C Error Formatting

dpgraham:
* Remembers protocol requested in createSession
    * Checks the parameters in 'createSession' and determines if it's W3C or MJSONWP
    * MJSONWP expects `{desiredCapabilities: {...}`; W3C expects `{capabilities: {alwaysMatch: {...}, firstMatch: [{...}, ...]}`
    * Also renamed 'processCaps' to 'processCapabilities'
* Add W3C error format
    * Add W3C status codes to errors
    * When deciding error format to return in mjsonwp.js, check format and then return eitherJSONWP or W3C format

(related to https://github.com/appium/appium/issues/9574)

### DIFF
--- a/lib/basedriver/driver.js
+++ b/lib/basedriver/driver.js
@@ -191,11 +191,11 @@ class BaseDriver extends MobileJsonWireProtocol {
   /**
    * Test createSession inputs to see if this is a W3C Session or a MJSONWP Session
    */
-  getProtocolFromCaps (desiredCapabilities, requiredCapabilities, capabilities) {
+  determineProtocol (desiredCapabilities, requiredCapabilities, capabilities) {
     if (_.isObject(capabilities) && _.isObject(capabilities.alwaysMatch)) {
-      return 'W3C';
+      return this.protocol = 'W3C';
     }
-    return 'MJSONWP';
+    this.protocol = 'MJSONWP';
   }
 
   // This is the main command handler for the driver. It wraps command
@@ -207,7 +207,7 @@ class BaseDriver extends MobileJsonWireProtocol {
     let startTime = Date.now();
     if (cmd === 'createSession') {
       // If creating a session determine if W3C or MJSONWP protocol was requested and remember the choice
-      this.protocol = this.getProtocolFromCaps(...args);
+      this.determineProtocol(...args);
       this.logEvent(EVENT_SESSION_INIT);
     } else if (cmd === 'deleteSession') {
       this.logEvent(EVENT_SESSION_QUIT_START);
--- a/lib/basedriver/driver.js
+++ b/lib/basedriver/driver.js
@@ -188,6 +188,16 @@ class BaseDriver extends MobileJsonWireProtocol {
     return true;
   }
 
+  /**
+   * Test createSession inputs to see if this is a W3C Session or a MJSONWP Session
+   */
+  getProtocolFromCaps (desiredCapabilities, requiredCapabilities, capabilities) {
+    if (_.isObject(capabilities) && _.isObject(capabilities.alwaysMatch)) {
+      return 'W3C';
+    }
+    return 'MJSONWP';
+  }
+
   // This is the main command handler for the driver. It wraps command
   // execution with timeout logic, checking that we have a valid session,
   // and ensuring that we execute commands one at a time. This method is called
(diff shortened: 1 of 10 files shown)


lib/basedriver/driver.js
```diff
@@ -188,6 +188,16 @@ class BaseDriver extends MobileJsonWireProtocol {
     return true;
   }
 
+  /**
+   * Test createSession inputs to see if this is a W3C Session or a MJSONWP Session
+   */
+  determineProtocol (desiredCapabilities, requiredCapabilities, capabilities) {
+    if (_.isObject(capabilities) && _.isObject(capabilities.alwaysMatch)) {
```
mykola-mokhnach: perhaps _isPlainObject_?

lib/basedriver/driver.js
```diff
@@ -188,6 +188,16 @@ class BaseDriver extends MobileJsonWireProtocol {
...
+      return this.protocol = 'W3C';
+    }
+    this.protocol = 'MJSONWP';
```
mykola-mokhnach: return?

mykola-mokhnach: should we set it to null in the class constructor?

lib/mjsonwp/errors.js
```diff
@@ -300,6 +363,17 @@ class NotImplementedError extends MJSONWPError {
   }
 }
 
+class UnableToCaptureScreen extends MJSONWPError {
+  static w3cStatus () {
+    return HTTPStatusCodes.INTERNAL_SERVER_ERROR;
+  }
+  constructor (err) {
+    super(err || 'Unable to capture screen', null, UnableToCaptureScreen.w3cStatus());
```
mykola-mokhnach: is  it ok to set jsonwp code to null?

lib/mjsonwp/mjsonwp.js
```diff
@@ -311,7 +319,11 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
       }
       // if anything goes wrong, figure out what our response should be
       // based on the type of error that we encountered
-      [httpStatus, httpResBody] = getResponseForJsonwpError(actualErr);
+      if (driver.protocol === 'W3C') {
```
mykola-mokhnach: should it be a constant?

lib/basedriver/driver.js
mykola-mokhnach: `return (_.isPlainObject(capabilities) && _.isPlainObject(capabilities.alwaysMatch)) ? W3C : MJSONWP`

imurchie: I would agree that this would be better as a pure function, that just returns the correct protocol. No longer needs to be called in the context of the driver (can be a utility method instead).

lib/mjsonwp/errors.js
```diff
@@ -1,31 +1,40 @@
...
 import { util } from 'appium-support';
+import HTTPStatusCodes from 'http-status-codes';
 
 // base error class for all of our errors
 class MJSONWPError extends ES6Error {
-  constructor (msg, jsonwpCode) {
+  constructor (msg, jsonwpCode, w3cStatus) {
     super(msg);
     this.jsonwpCode = jsonwpCode;
+    this.w3cStatus = w3cStatus || 400; // Most w3c errors are 400 error so make that the default
```
imurchie: Why not set this as a default in the function declaration?
```js
constructor (msg, jsonwpCode, w3cStatus = 400) {
```

lib/mjsonwp/mjsonwp.js
```diff
@@ -189,6 +189,14 @@ function getResponseForJsonwpError (err) {
   return [httpStatus, httpResBody];
 }
 
+function getResponseForW3CError (err) {
+  let httpStatus = err.w3cStatus || 400;
```
imurchie: `400` seems to be known in too many places. Only the error module should know it, really. It would be nicer to either have a guarantee that `err.w3cStatus` is set (could make it a getter on the base error object to ensure that something is returned if nothing is there when it is accessed), or as a constant exported from the error module.

lib/mjsonwp/errors.js
dpgraham: I'll set it to 13 by default if null is provided. 13 is just an `Unknown Error`

dpgraham:
Changes pushed

PR #146, opened 2017-11-07
Translate W3C touch actions to multi actions

dpgraham:
(related to https://github.com/appium/appium/issues/9579)


lib/basedriver/commands/actions.js
```diff
@@ -0,0 +1,117 @@
...
+    throw new errors.NotYetImplementedError(`Could not translate W3C actions ${JSON.stringify(actions)} to JSONWP actions. 'performMultiAction' command not yet implemented`);
+  }
+
+  log.debug(`Translated ${JSON.stringify(actions)} to ${JSON.stringify(multiActions)}`);
+  log.debug(`Calling 'performMultiAction' with translated actions`);
+  return await this.performMultiAction(multiActions);
+};
+
+const ACTION_TYPES = {
+  'pointer': 'pointer',
```
mykola-mokhnach: we don't need quotes for key names

lib/basedriver/commands/actions.js
```diff
@@ -0,0 +1,117 @@
...
+
+let commands = {}, helpers = {}, extensions = {};
+
+/**
+ * If w3c performActions was called, attempt to translate it to JSONWP actions, if possible.
+ */
+commands.performActions = async function (actions) {
+  let multiActions;
+  log.debug(`Attempting to translate W3C actions to JSONWP actions`);
+  multiActions = helpers.translateActions(actions);
```
mykola-mokhnach: do we really need to do the translation? there is the separate /actions endpoint now. So let's implement the W3C protocol for it in each client lib and keep the old endpoints for JSONWP format. There are too many possible different chain variations, that we physically cannot cover with tests.

dpgraham: If you implement it in a driver, this command will be overridden. I just figure it's better than nothing. What do you think @imurchie?

dpgraham:
@imurchie What do you think about this? Is this a worthwhile pull request or should we just do it all per driver?

lib/basedriver/commands/actions.js
mykola-mokhnach: my point is that it will be necessary for us to support one more abstraction layer. I'd rather keep it simple and let Appium just pass the request body to the server as is. So then the server provider could fix it on his side if there are issues.

lib/basedriver/commands/index.js
```diff
@@ -9,7 +9,7 @@ Object.assign(
   sessionCmds,
   settingsCmds,
   timeoutCmds,
-  findCmds
+  findCmds,
```
mykola-mokhnach: is this comma needed?

lib/mjsonwp/validators.js
```diff
@@ -14,7 +14,7 @@ function msValidator (ms) {
 const validators = {
   setUrl: (url) => {
     // either an `xyz://`, `about:`, or `data:` scheme is allowed
-    if (!url || !url.match(/^([a-zA-Z0-9_-]+:\/\/)|(about:)|(data:)/)) {
+    if (!url || !url.match(/^([a-zA-Z0-9_+.-]+:\/\/)|(about:)|(data:)/)) {
```
mykola-mokhnach: I assume this was a part of some other PR

test/mjsonwp/validator-specs.js
```diff
@@ -17,16 +17,19 @@ describe('MJSONWP', () => {
         (() => {validators.setUrl('foo');}).should.throw(/url/i);
       });
       it('should succeed when given url starting with http', async () => {
-        (() => {validators.setUrl('http://appium.io');}).should.not.throw;
+        (() => {validators.setUrl('http://appium.io');}).should.not.throw();
```
mykola-mokhnach: seems like merge issue %)

lib/basedriver/commands/index.js
dpgraham: No but it doesn't hurt. It's actually pretty common practice and some linters even require a dangling comma for line separated lists.

test/mjsonwp/validator-specs.js
dpgraham: Fixed it. The W3C branch was behind master. Rebased and did a force push.

PR #147, opened 2017-11-07
w3c routes

dpgraham:
* Some of the routes just point to the an existing command
* Implemented the 'timeouts' endpoint
* The rest of them are just new routes with different command names that are to be implemented by drivers (most of them are web commands)

### DIFF
--- a/lib/basedriver/commands/timeout.js
+++ b/lib/basedriver/commands/timeout.js
@@ -24,6 +24,13 @@ commands.timeouts = async function (type, duration) {
   }
 };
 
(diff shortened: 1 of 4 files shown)


imurchie:
For clarity's sake, we had started a new section in the routes for the W3C stuff, and what it replaces: https://github.com/appium/appium-base-driver/blob/f7b4e818f882da5b9bba1b5e0dee09d6e066107c/lib/mjsonwp/routes.js#L494

dpgraham:
Changes made

PR #148, opened 2017-11-09
Add image-helper with methods for cropping images

astappev:
See [previous thread](https://github.com/appium/appium-base-driver/pull/144) for more details


imurchie:
I would rather see this with the rest of the supporting functions that are used in multiple places, inside `appium-support`.

danielputerman:
@imurchie So would you like us to remove the method from the base-driver? Or is it possible to accept the PR in its current state and then refactor the method out into the appium-support?

mykola-mokhnach:
@danielputerman I think @imurchie would like to see this helper in https://github.com/appium/appium-support/tree/master/lib. Just create a new module there, call it image-util.js and put this code there.

imurchie:
Yes. I would rather not have to refactor things if possible.

PR #149, opened 2017-11-10
Added route for setDeviceTime

anthonypdawson:
Implementing setting device time for the Android platform across this repo, appium-android-driver and python-client

### DIFF
--- a/lib/mjsonwp/routes.js
+++ b/lib/mjsonwp/routes.js
@@ -302,7 +302,8 @@ const METHOD_MAP = {
     POST: {command: 'mobileShake'}
   },
   '/wd/hub/session/:sessionId/appium/device/system_time': {
-    GET: {command: 'getDeviceTime'}
+    GET: {command: 'getDeviceTime'},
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
This endpoint won't work on iOS anyway. Also, it makes less sense on other platforms. I'd rather add as a apart of `mobile: ` interface for Android only.

mykola-mokhnach:
@jlipps @imurchie What do you think?

mykola-mokhnach:
GET is ok, since it more or less available everywhere, but setTime is blocked on many platforms

imurchie:
I agree this could be a setting, or a desired capability even. What is the use case for changing the device date/time mid-session?

mykola-mokhnach:
This PR has been superseeded by https://github.com/mykola-mokhnach/appium/blob/9f36377d7ba47d7b55ae4d98577872917e6f083d/docs/en/writing-running-appium/android/android-shell.md

PR #150, opened 2017-11-11
Add broadcast intent endpoint

dpgraham:
* Endpoint is `/device/broadcast_intent`
* Takes 'intent' as JSON argument

(related to https://github.com/appium/appium/issues/3172)

### DIFF
--- a/lib/mjsonwp/routes.js
+++ b/lib/mjsonwp/routes.js
@@ -369,6 +369,9 @@ const METHOD_MAP = {
   '/wd/hub/session/:sessionId/appium/device/current_package': {
     GET: {command: 'getCurrentPackage'}
   },
+  '/wd/hub/session/:sessionId/appium/device/broadcast_intent': {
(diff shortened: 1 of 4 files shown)


lib/mjsonwp/routes.js
```diff
@@ -369,6 +369,9 @@ const METHOD_MAP = {
   '/wd/hub/session/:sessionId/appium/device/current_package': {
     GET: {command: 'getCurrentPackage'}
   },
+  '/wd/hub/session/:sessionId/appium/device/broadcast_intent': {
```
mykola-mokhnach: this is android-specific endpoint. Wouldn't be more appropriate to implement it via executeScript?

imurchie: The execute "endpoints" were meant as temporary stop-gaps. There is no reason to make an Android-specific option map there.

mykola-mokhnach:
Can we close this PR now @dpgraham ?

dpgraham:
Yep.

PR #151, opened 2017-11-15
Allow dot and plus character in valid uri scheme

studuffy:
The main reason for this pull request was to fix the URI scheme validator, to ensure dot and plus characters are recognised as allowed, in line with [rfc 2396](http://www.ietf.org/rfc/rfc2396.txt).

In running the tests for this, it seemed that `.should.not.throw` in the validator specs was not actually checking anything and was always passing even if something was thrown, and instead should be `.should.not.throw()`, so I've changed that too.


studuffy:
It seems my email wasn't associated with my commits so the CLA check was failing. Fixed now!

imurchie:
Published in `appium-base-driver@2.15.8`.

PR #152, opened 2017-11-21
W3C

dpgraham:
These commits have already been reviewed. The W3C pull requests were all made to a separate W3C branch (not master) because I wanted the changes to all be merged to master at the same time so we didn't have a partial W3C implementation.

(https://github.com/appium/appium/issues/9579)


dpgraham:
Published to appium-base-driver 2.16.0

PR #154, opened 2017-12-15
Support pre-W3C element screenshot endpoint

mykola-mokhnach:
The new W3C-compatible endpoint is only available since Selenium 3.8.0 (which is currently not supported even by beta version of java client), although I think it costs nothing for us to support the legacy endpoint as well.


imurchie:
Published in `appium-base-driver@2.17.1`.

PR #157, opened 2017-12-19
Fix incorrect identifying of unprefixed caps

dpgraham:
* Appium logs warnings about W3C caps that aren't prefixed with `appium:`
* Was incorrectly warning about prefixed caps.
* Fix makes it so it warns about unprefixed, non-standard caps.

### DIFF
--- a/lib/basedriver/capabilities.js
+++ b/lib/basedriver/capabilities.js
@@ -76,7 +76,7 @@ function stripAppiumPrefixes (caps) {
   const prefixedCaps = _.filter(_.keys(caps), cap => `${cap}`.startsWith(prefix));
   const badPrefixedCaps = [];
   const unprefixedCaps = _.filter(_.keys(caps), (cap) => (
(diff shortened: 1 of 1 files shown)


lib/basedriver/capabilities.js
```diff
@@ -76,7 +76,7 @@ function stripAppiumPrefixes (caps) {
   const prefixedCaps = _.filter(_.keys(caps), cap => `${cap}`.startsWith(prefix));
   const badPrefixedCaps = [];
   const unprefixedCaps = _.filter(_.keys(caps), (cap) => (
-    `${cap}`.includes(':') && !isStandardCap(cap)
```
imurchie: I don't think `_.keys` will ever return anything but an array of strings (it relies on the logic of http://ecma-international.org/ecma-262/7.0/#sec-enumerableownnames) so the coercing it to string is not needed.

dpgraham:
Fixes pushed.

dpgraham:
Published to 2.18.1

PR #158, opened 2017-12-20
Make sure that temporary files are cleaned up properly while configuring the app

mykola-mokhnach:
The current code leaves too many file leftovers behind if the downloaded stuff requires unzipping or some failure happens during the extraction process.

### DIFF
--- a/lib/basedriver/helpers.js
+++ b/lib/basedriver/helpers.js
@@ -21,25 +21,25 @@ async function configureApp (app, appExt, mountRoot="Volumes", windowsShareUserN
   let shouldUnzipApp = false;
 
   // check if we're copying from a windows network share
-  if (app.startsWith("\\")) {
-    app = await copyFromWindowsNetworkShare(app, appExt,
+  if (newApp.startsWith("\\")) {
+    newApp = await copyFromWindowsNetworkShare(newApp, appExt,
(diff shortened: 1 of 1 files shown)


lib/basedriver/helpers.js
```diff
@@ -12,38 +12,51 @@ const ZIP_EXTS = ['.zip', '.ipa'];
...
   }
 
-  let newApp = null;
+  let newApp = app;
   let shouldUnzipApp = false;
 
   // check if we're copying from a windows network share
-  if (_.startsWith(app, "\\")) {
-    app = await copyFromWindowsNetworkShare(app, appExt, mountRoot, windowsShareUserName, windowsSharePassword);
+  if (newApp.startsWith("\\")) {
```
dpgraham: Does the native `startsWith` work with all the versions of Node that we support?

http://node.green/#ES2015-built-ins-well-known-symbols-Symbol-match--String-prototype-startsWith

mykola-mokhnach: I assume babel should handle it

dpgraham: @mykola-mokhnach Maybe. I'll check to confirm.

dpgraham: Yeah, the polyfill covers it.

PR #159, opened 2017-12-20
Don't validate 'presence' on firstMatch caps

dpgraham:
* Doesn't make sense to validate 'presence' in the firstMatch caps, because the firstMatch caps extend alwaysMatch and is going to miss necessary capabilities
* Only 'alwaysMatch' should validate 'presence'
* Simply set the 'skipPresenceConstraint' in 'validateCaps' to true for firstMatch caps


dpgraham:
Published to 2.18.2

PR #164, opened 2018-01-07
Fix implicit wait for w3c

KazuCocoa:
According to [the W3C spec](https://github.com/jlipps/simple-wd-spec#set-timeouts), clients send `{implicit: 3000}` as request body to the server. But current implementation expects `{type: implicit, ms: 3000}` for both W3C and MJSONWP.

I've changed to `optional` parameter because of supporting both W3C format and MJSONWP one.
https://github.com/appium/appium-base-driver/compare/master...KazuCocoa:fix_timeout_for_w3c?expand=1#diff-28707e6fe74fc702b56e27eaff233df0R25

Ideally, I'd like to remain `required: ['type', 'ms']}` for MJSONWP and set `optional: ['script', 'pageLoad', 'implicit']` for W3C, but I didn't come up with good idea... Do you have any good way to implement this?

# Server log when  Ruby client sends set-implicit-wait-time requests to the server
## W3C format
```
[debug] [JSONWP Proxy] Proxying [POST /session] to [POST http://localhost:8100/session] with body: {"desiredCapabilities":{"bundleId":"com.example.apple-samplecode.UICatalog","arguments":[],"environment":{},"shouldWaitForQuiescence":true,"shouldUseTestManagerForVisibilityDetection":false,"maxTypingFrequency":60,"shouldUseSingletonTestManager":true}}
[debug] [JSONWP Proxy] Got response with status 200: {"value":{"sessionId":"94BC124C-0414-43F8-A94A-AF633EBE48F0","capabilities":{"device":"iphone","browserName":"UICatalog","sdkVersion":"10.3.1","CFBundleIdentifier":"com.example.apple-samplecode.UICatalog"}},"sessionId":"94BC124C-0414-43F8-A94A-AF633EBE48F0","status":0}
[debug] [BaseDriver] Event 'wdaSessionStarted' logged at 1515329565137 (21:52:45 GMT+0900 (JST))
[debug] [BaseDriver] Event 'wdaStarted' logged at 1515329565138 (21:52:45 GMT+0900 (JST))
[XCUITest] Skipping setting of the initial display orientation. Set the "orientation" capability to either "LANDSCAPE" or "PORTRAIT", if this is an undesired behavior.
[debug] [BaseDriver] Event 'orientationSet' logged at 1515329565138 (21:52:45 GMT+0900 (JST))
[Appium] New XCUITestDriver session created successfully, session 70e94037-b9e6-494b-afdf-0fe99c3d9f54 added to master session list
[...]

PR #165, opened 2018-01-08
W3C remediations

dpgraham:
* Fixed a defect in 'firstMatch' validation
  * Previously, it was validating the firstMatch capabilities and was failing on the first validation failure, which is a problem if firstMatch has a length > 1 and it fails before finding a potentially passing capabilities set is found
  * Now, it doesn't error-out on the first failure. If it fails, it skips over it to the next one until it finds one that passes and returns that.

* Altered the 'presence' constraint validation
  *  Don't validate 'presence' on 'alwaysMatch' because if a capability isn't present on 'alwaysMatch' it may be present in one of the 'firstMatch' capabilities
  *  Validate presence on all of the 'firstMatch' capabilities, except for the keys that are present in 'alwaysMatch'. If the key is in 'alwaysMatch', strip out the presence constraint because we know that it has passed already

* Changed the default protocol
  * If MJSONWP _and_ W3C are provided, initiate a W3C session by default. If clients are passing in W3C parameters to create a session, we can safely assume that they've implemented the rest of the W3C spec

* Improved error logging
  * If errors were found, show a message saying `Could not find matching capabilities: ${errors}` where `errors` are a list of all the validation error messages for each of the attempts to validate `firstMatch` caps
  * If `firstMatch` caps has a length of one, then don't show the message saying `Could not find matching capabilities` portion of it, just show the errors

* Other remediations
  * Allow `alwaysMatch` and `firstMatch` to be `null` or `undefined`
  * Added single quotes around names of capabilities when showing their errors


lib/basedriver/capabilities.js
```diff
@@ -11,7 +11,7 @@ function mergeCaps (primary = {}, secondary = {}) {
   for (let [name, value] of _.toPairs(secondary)) {
     // Overwriting is not allowed. Primary and secondary must have different properties (w3c rule 4.4)
     if (!_.isUndefined(primary[name])) {
-      throw new Error(`property ${name} should not exist on both primary and secondary`);
+      throw new Error(`property '${name}' should not exist on both primary and secondary`);
```
mykola-mokhnach: I think we can improve his error message by actually displaying both primary and secondary objects via JSON.stringify

lib/basedriver/capabilities.js
```diff
@@ -20,16 +20,19 @@ function mergeCaps (primary = {}, secondary = {}) {
 }
 
 // Validates caps against a set of constraints
-function validateCaps (caps, constraints = {}, skipPresenceConstraint = false) {
+function validateCaps (caps, constraints = {}, opts = {}) {
```
mykola-mokhnach: Can you please sync your changes with @KazuCocoa (https://github.com/appium/appium-base-driver/pull/164)

lib/basedriver/capabilities.js
```diff
@@ -42,10 +45,10 @@ function validateCaps (caps, constraints = {}, skipPresenceConstraint = false) {
     let message = [];
     for (let [attribute, reasons] of _.toPairs(validationErrors)) {
       for (let reason of reasons) {
-        message.push(` ${attribute} ${reason},`);
+        message.push(`'${attribute}' ${reason}`);
       }
     }
-    throw new Error(message.join(''));
+    throw new Error(message.join(', '));
```
mykola-mokhnach: perhaps ';' ?

lib/basedriver/capabilities.js
```diff
@@ -106,40 +109,55 @@ function stripAppiumPrefixes (caps) {
...
   if (shouldValidateCaps) {
-    requiredCaps = validateCaps(requiredCaps, constraints);
+    requiredCaps = validateCaps(requiredCaps, constraints, {skipPresenceConstraint: true});
   }
 
-  // Validate all of the first match capabilities (see spec #5)
+
+  // Remove the 'presence' constraint for any keys that are already present in 'requiredCaps'
+  // since we know that this constraint has already passed
+  let constraintsCopy = _.clone(constraints);
```
mykola-mokhnach: I'd rather use .filter on it ;)

lib/basedriver/capabilities.js
```diff
@@ -152,16 +170,25 @@ function parseCaps (caps, constraints = {}, shouldValidateCaps = true) {
...
-  return {requiredCaps, allFirstMatchCaps, validatedFirstMatchCaps, matchedCaps};
+  return {requiredCaps, allFirstMatchCaps, validatedFirstMatchCaps, matchedCaps, errors};
 }
 
 // Calls parseCaps and just returns the matchedCaps variable
 function processCapabilities (caps, constraints = {}, shouldValidateCaps = true) {
-  return parseCaps(caps, constraints, shouldValidateCaps).matchedCaps;
+  const {matchedCaps, errors} = parseCaps(caps, constraints, shouldValidateCaps);
+
+  if (matchedCaps === null) {
```
mykola-mokhnach: _.isNull

lib/basedriver/capabilities.js
```diff
@@ -152,16 +170,25 @@ function parseCaps (caps, constraints = {}, shouldValidateCaps = true) {
...
+    if (_.isArray(caps.firstMatch) && caps.firstMatch.length > 1) {
+      // If there was more than one 'firstMatch' cap, indicate that we couldn't find a matching capabilities set and show all the errors
+      throw new Error(`Could not find matching capabilities from ${JSON.stringify(caps)}:\n ${errors.join('\n')}`);
+    } else {
+      // Otherwise, just show the singular error message
+      throw new Error(errors[0]);
```
mykola-mokhnach: should we check if errors is not empty first?

lib/basedriver/commands/session.js
```diff
@@ -13,10 +13,10 @@ commands.createSession = async function (desiredCapabilities, requiredCaps, capa
   }
 
   let caps;
```
mykola-mokhnach:
```
let caps = desiredCapabilities
if (capabilities) {
  caps = processCapabilities...
}
```

lib/basedriver/driver.js
```diff
@@ -199,7 +199,7 @@ class BaseDriver extends MobileJsonWireProtocol {
    * Test createSession inputs to see if this is a W3C Session or a MJSONWP Session
    */
   static determineProtocol (desiredCapabilities, requiredCapabilities, capabilities) {
-    return (_.isPlainObject(capabilities) && _.isPlainObject(capabilities.alwaysMatch)) ?
+    return (_.isPlainObject(capabilities)) ?
```
mykola-mokhnach: parentheses are not needed here

lib/basedriver/capabilities.js
dpgraham: It can't be empty. If it found an error in a singleton `firstMatch` array then there was exactly one error.

lib/basedriver/capabilities.js
dpgraham: That doesn't actually work. I tried doing it the 'lodash' way but the problem is it needs to delete just one property of the constraint object and not the whole thing. So something like this....

```javascript
{
  foo: {
    presence: true,
    required: true,
  }
}
```

Needs to be translated to this (assuming that 'foo' presence isn't required anymore)
```javascript
{
  foo: {
    required: true
  }
}
```

lib/basedriver/capabilities.js
dpgraham: Yeah, good call. The messages may have commas in them and that would make things confusing.

lib/basedriver/capabilities.js
dpgraham: Yeah, once we approve #164 I can take care of rebasing.

lib/basedriver/capabilities.js
mykola-mokhnach: How about this?

```js
const constraintsCopy = _.toPairs(constraints)
    .filter((x) => !_.keys(requiredCaps).includes(x[0]))
    .reduce((acc, value) => {
       acc[value[0]] = value[1];
       return acc;
    }, {});
```

lib/basedriver/capabilities.js
mykola-mokhnach: I would ask you to take a look at that PR, so you know what is going to be changed

lib/basedriver/capabilities.js
mykola-mokhnach: also I assume the name of `constraintsCopy` variable itself might be more specific

lib/basedriver/capabilities.js
dpgraham: Yeah I've looked at it and added some comments. Should be a pretty simple fix.

lib/basedriver/capabilities.js
dpgraham: Yeah I'll rename it `filterredConstraints`. Although I generally agree that I like using functional programming patterns over iterative, I think this may be a case where the iterative way is more clear.

lib/basedriver/capabilities.js
```diff
@@ -125,6 +125,11 @@ function parseCaps (caps, constraints = {}, shouldValidateCaps = true) {
     throw new Error('The capabilities.firstMatch argument was not valid for the following reason(s): "capabilities.firstMatch" must be a JSON array or undefined');
   }
 
+  // If an empty array as provided, we'll be forgiving and make it an array of one empty object
+  if (allFirstMatchCaps.length === 0) {
```
mykola-mokhnach: you could use _.isEmpty

lib/basedriver/capabilities.js
```diff
@@ -48,7 +48,7 @@ function validateCaps (caps, constraints = {}, opts = {}) {
         message.push(`'${attribute}' ${reason}`);
       }
     }
-    throw new Error(message.join(', '));
+    throw new Error(message.join('; '));
```
mykola-mokhnach: I'd rather rename it to messages or errorMessages

dpgraham:
@jlipps @imurchie will either of you have time to look at this?

lib/basedriver/capabilities.js
```diff
@@ -106,40 +109,60 @@ function stripAppiumPrefixes (caps) {
...
   if (shouldValidateCaps) {
-    requiredCaps = validateCaps(requiredCaps, constraints);
+    requiredCaps = validateCaps(requiredCaps, constraints, {skipPresenceConstraint: true});
   }
 
-  // Validate all of the first match capabilities (see spec #5)
+
+  // Remove the 'presence' constraint for any keys that are already present in 'requiredCaps'
+  // since we know that this constraint has already passed
+  let filterredConstraints = {...constraints};
```
jlipps: `filtered` (one r)

lib/basedriver/commands/session.js
```diff
@@ -12,12 +12,10 @@ commands.createSession = async function (desiredCapabilities, requiredCaps, capa
...
   }
 
-  let caps;
-  if (desiredCapabilities) {
-    caps = desiredCapabilities;
-  } else {
-    caps = processCapabilities(capabilities, this.desiredCapConstraints, this.shouldValidateCaps);
-  }
+  // If we were provided with W3C capabilities, parse those. Otherwise, fall back to MJSONWP.
+  let caps = capabilities ?
```
jlipps: i know the variables for this method were set before this PR, but coming to this code for the first time, it would have been nicer to see them named something like: `mjsonwpDesiredCapabilities`, `mjsonwpRequiredCaps`, `w3cCapabilities`.

dpgraham: Yeah that's a good point. My next incoming PR for base-driver renames a lot of those.

PR #166, opened 2018-01-10
Fix problems with W3C responses

dpgraham:
* Make malformed caps return 400 error  
* Add InvalidArgumentTypeError to errors
* Made 'NotYetImplementedError' a 405 (Method Not Allowed) status code
* Made errors in 'createSession' return InvalidArgumentTypeError when capabilities are invalid or malformed
* Fix incorrect W3C responses 
    * Add w3c status for NotImplemented and NotYetImplemented
    * Changed `getResponseForW3CStatusError` so that it throws an unknown error if the exception isn't a w3c exception
      * Checks for property `.w3cStatus` on the Error to confirm if it's a proper W3C protocol-blessed exception (@jlipps do you think this is adequate?)
    * Moved logic that reassigned errors into MJSONWP conditional blocks; W3C shouldn't reassign errors because they all have their own specific HTTP status codes
    * Added endpoint tests that check that you can create sessions in W3C and that subsequent calls to the session return the correct responses

### DIFF
--- a/lib/basedriver/capabilities.js
+++ b/lib/basedriver/capabilities.js
@@ -2,6 +2,7 @@ import _ from 'lodash';
 import { validator } from './desired-caps';
 import { util } from 'appium-support';
 import log from './logger';
+import { errors } from '../mjsonwp/errors';
 
 // Takes primary caps object and merges it into a secondary caps object.
 // (see https://www.w3.org/TR/webdriver/#dfn-merging-capabilities)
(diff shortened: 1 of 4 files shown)


lib/basedriver/capabilities.js
```diff
@@ -25,7 +26,7 @@ function validateCaps (caps, constraints = {}, opts = {}) {
   let  {skipPresenceConstraint} = opts;
 
   if (!_.isObject(caps)) {
```
mykola-mokhnach: should it be isPlainObject instead?

lib/mjsonwp/mjsonwp.js
```diff
@@ -191,7 +191,12 @@ function getResponseForJsonwpError (err) {
 }
 
 function getResponseForW3CError (err) {
-  let httpStatus = err.w3cStatus;
+  let httpStatus;
+  if (!(err.w3cStatus)) {
```
mykola-mokhnach: the internal parentheses are not needed

lib/mjsonwp/mjsonwp.js
```diff
@@ -191,7 +191,12 @@ function getResponseForJsonwpError (err) {
...
+    log.error(`Encountered internal error running command: ${err.stack}`);
+    err = new errors.UnknownError(err.message);
+  }
+  httpStatus = err.w3cStatus;
```
mykola-mokhnach: is it ok if httpStatus is undefined?

lib/basedriver/capabilities.js
dpgraham: I may as well change it to `isPlainObject` although because it's parsed from JSON it couldn't be a non-plain object.

dpgraham:
Fixes pushed

PR #167, opened 2018-01-10
Set automations in one file

dpgraham:
* Made an 'automations.js' file that lists all valid automations as an Object instead of hard-coding it
* Make desired-caps pull in the values of those automations for validation instead of hard-coding them
* Expose 'validAutomations' to Driver object
* appium/appium will use this instead of hardcoding automation names to avoid possible errors

### DIFF
--- a/lib/basedriver/automations.js
+++ b/lib/basedriver/automations.js
@@ -1,0 +1,12 @@
+// Lists all the names of the automations that Appium supports
+const automations = {
+  APPIUM: 'Appium',
+  SELENDROID: 'Selendroid',
+  UIAUTOMATOR2: 'UiAutomator2',
+  XCUITEST: 'XCUITest',
+  YOUIENGINE: 'YouiEngine',
+  ESPRESSO: 'Espresso',
+  FAKE: 'Fake',
+};
+
+export { automations };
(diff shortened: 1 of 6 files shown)


lib/basedriver/automations.js
```diff
@@ -0,0 +1,12 @@
+// Lists all the names of the automations that Appium supports
+const automations = {
```
jlipps: would prefer `automationName` throughout instead of `automation`

mykola-mokhnach: how about Windows and Mac OS drivers?

imurchie: Windows and Mac drivers, along with android and ios, are handled by `platformName` not `automationName` (see https://github.com/appium/appium/blob/master/lib/appium.js#L62-L127).

mykola-mokhnach:
Can we also fix the tests?

jlipps:
for what it's worth, in appium 2 what i'd like to see is:

- basedriver knows nothing about automationNames
- the main appium package detects which drivers are installed
- each driver reports its own automation name
- appium then uses this info to determine the valid automation names for a session

that way we're not keeping knowledge about platform-specific stuff in the more general code

dpgraham:
@jlipps I agree. We wouldn't need to 'validate' the automation name in 2.0; we would just need have it look for the driver and then respond with an error if it couldn't find one associated with the automation name.

PR #170, opened 2018-01-12
Fix broken Proxy logic for W3C

dpgraham:
* Moved 'ProxyRequestError' handling outside of the MJSONWP block.
  * When it's a proxy error, Appium translates the error (based on statusCode) to it's proper Error
  * W3C sessions weren't doing this when ProxyRequestError was being used
  * Moved conditionals so that W3C translates Proxy errors too
* Added tests in `mjsonwp-e2e-specs` that test JWProxy
  * Made a dummy express server that is used as a jwproxy server that binds routes dynamically
  * Test successful 200 proxies (this test passed _without_ coding fix)
  * Tested unsuccessful responses to see that a JSONWP response with status = 6 would return the correct 'NoSuchDriverError' with statusCode = 404

### DIFF
--- a/lib/mjsonwp/mjsonwp.js
+++ b/lib/mjsonwp/mjsonwp.js
@@ -324,19 +324,17 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
       // if anything goes wrong, figure out what our response should be
       // based on the type of error that we encountered
       let actualErr = err;
-      if (driver.protocol === BaseDriver.DRIVER_PROTOCOL.MJSONWP) {
-        if (isErrorType(err, errors.ProxyRequestError)) {
-          log.error(`Encountered internal error running command:  ${JSON.stringify(err)} ${err.stack}`);
-          actualErr = err.getActualError();
-        } else if (!(isErrorType(err, MJSONWPError) ||
-          isErrorType(err, errors.BadParametersError))) {
-          log.error(`Encountered internal error running command: ${err.stack}`);
-          actualErr = new errors.UnknownError(err);
-        }
-        [httpStatus, httpResBody] = getResponseForJsonwpError(actualErr);
-      } else {
-        [httpStatus, httpResBody] = getResponseForW3CError(actualErr);
+
+      // If it's a JWProxy error, translate it to a MJSONWPError exception
+      if (isErrorType(err, errors.ProxyRequestError)) {
+        log.error(`Encountered internal error running command:  ${JSON.stringify(err)} ${err.stack}`);
+        actualErr = err.getActualError(driver.protocol);
+      } else if (driver.protocol === BaseDriver.DRIVER_PROTOCOL.MJSONWP && (!(isErrorType(err, MJSONWPError) || isErrorType(err, errors.BadParametersError)))) {
+        log.error(`Encountered internal error running command: ${err.stack}`);
+        actualErr = new errors.UnknownError(err);
       }
+
+      [httpStatus, httpResBody] = driver.protocol === BaseDriver.DRIVER_PROTOCOL.MJSONWP ? getResponseForJsonwpError(actualErr) : getResponseForW3CError(actualErr);
     }
 
     // decode the response, which is either a string or json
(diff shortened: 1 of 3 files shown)


lib/mjsonwp/mjsonwp.js
```diff
@@ -324,19 +324,17 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
...
-        }
-        [httpStatus, httpResBody] = getResponseForJsonwpError(actualErr);
-      } else {
-        [httpStatus, httpResBody] = getResponseForW3CError(actualErr);
+
+      // If it's a JWProxy error, translate it to a MJSONWPError exception
+      if (isErrorType(err, errors.ProxyRequestError)) {
+        log.error(`Encountered internal error running command:  ${JSON.stringify(err)} ${err.stack}`);
+        actualErr = err.getActualError(driver.protocol);
+      } else if (driver.protocol === BaseDriver.DRIVER_PROTOCOL.MJSONWP && (!(isErrorType(err, MJSONWPError) || isErrorType(err, errors.BadParametersError)))) {
```
imurchie: `driver.protocol === BaseDriver.DRIVER_PROTOCOL.MJSONWP` is pretty unwieldy, and is used in a number of places. Could it be extracted into helper functions (`driver.isMjsonwpProtocol()` and `driver.isW3CProtocol()` or something)? Also limits the number of places that know about BaseDriver's internals

mykola-mokhnach: can we wrap these long lines?

jlipps: line seems really long now

test/mjsonwp/helpers.js
```diff
@@ -0,0 +1,25 @@
...
+export function createProxyServer (sessionId, port) {
+  // Start an express server for proxying
+  let app = new Express();
+  app.use(bodyParser.json());
+  let server = app.listen(port);
+  return {app, server};
+}
+
+let handlers = {
+  'post': {},
```
mykola-mokhnach: quotes are not needed

test/mjsonwp/helpers.js
```diff
@@ -0,0 +1,25 @@
...
+
+export function createProxyServer (sessionId, port) {
+  // Start an express server for proxying
+  let app = new Express();
+  app.use(bodyParser.json());
+  let server = app.listen(port);
+  return {app, server};
+}
+
+let handlers = {
```
jlipps: don't need quotes around keys

test/mjsonwp/mjsonwp-e2e-specs.js
```diff
@@ -467,15 +471,71 @@ describe('MJSONWP', async () => {
...
+            jwproxy.sessionId = sessionId;
+            driver.performActions = async (actions) => await jwproxy.command('/perform-actions', 'POST', actions);
+          });
+
+          afterEach(function () {
+            server.close();
+            delete driver.performActions;
+          });
+
+          it('should work if a proxied request returns a response with status 200', async () => {
```
mykola-mokhnach: async function

mykola-mokhnach: same below

dpgraham: They have `await` in them. If you're talking about `addHandler` the callback actually isn't async. It's an Express handler.

jlipps: he means that in mocha tests we should try to write `async function () {` instead of `async () => {`, since mocha needs binding to work properly. it usually doesn't matter and i'm on the fence about making this a requirement, but it's not a bad idea.

dpgraham:
Fixes made.... I had to change fake-driver to extend BaseDriver instead of MJSONWP. What does that MJSONWP class even do?

/lib/mjson.wp line 14:
```javascript

class MJSONWP {}
```

test/mjsonwp/mjsonwp-e2e-specs.js
dpgraham: Aha, yeah it probably should be a requirement. That's bitten me in the past a few times.

dpgraham: I think there's an ESLint rule for that.

dpgraham:
@jlipps @mykola-mokhnach @imurchie Can i merge this?

test/mjsonwp/mjsonwp-e2e-specs.js
imurchie: https://github.com/appium/eslint-config-appium/pull/12

PR #171, opened 2018-01-15
Check that W3C capabilities is a plain object

dpgraham:
* If 'capabilities' and 'desiredCapabilities' were provided, and 'capabilities' is NOT a plain object, fall back to 'desiredCapabilities'
* Added more logging to the W3C vs. MJSONWP determination logic
* Rename `createSession` args to be more descriptive
* Added basic W3C unit tests for above fix

### DIFF
--- a/test/basedriver/capability-specs.js
+++ b/test/basedriver/capability-specs.js
@@ -339,4 +339,36 @@ describe('Desired Capabilities', () => {
     }).should.eventually.be.rejectedWith(/blank/);
 
   });
+
+  describe('w3c', async () => {
+    it('should accept w3c capabilities', async () => {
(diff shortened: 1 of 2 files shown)


test/basedriver/capability-specs.js
```diff
@@ -339,4 +339,36 @@ describe('Desired Capabilities', () => {
     }).should.eventually.be.rejectedWith(/blank/);
 
   });
+
+  describe('w3c', async () => {
+    it('should accept w3c capabilities', async () => {
+      const [sessionId, caps] = await d.createSession(null, null, {
+        alwaysMatch: {
+          'platformName': 'iOS',
```
jlipps: don't need strings around alpha keys

PR #172, opened 2018-01-16
Assume a file is located on SMB server if the path starts with double backslash

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/10003


imurchie:
Seems like a test (https://github.com/appium/appium-base-driver/blob/3bffc168b8e17ee077ec31dc6a3ab92da1b7c7da/test/basedriver/helpers-e2e-specs.js#L50-L53) needs to be updated.

mykola-mokhnach:
thanks @imurchie 
The test is updated

PR #174, opened 2018-01-17
Consolidate non-prefixed caps warnings

dpgraham:
* Previously, it was going through alwaysMatch and all of the firstMatch caps and individually logging when any of them had unprefixed, non-standard caps
* Now it does them all at once
  * Made a helper function that takes a caps object ({alwaysMatch, firstMatch}) and finds all of the non-prefixed caps in all of them
  * Calls that function and logs warning

### DIFF
--- a/lib/basedriver/capabilities.js
+++ b/lib/basedriver/capabilities.js
@@ -79,9 +79,6 @@ function stripAppiumPrefixes (caps) {
   const prefix = 'appium:';
   const prefixedCaps = _.filter(_.keys(caps), cap => `${cap}`.startsWith(prefix));
   const badPrefixedCaps = [];
-  const unprefixedCaps = _.filter(_.keys(caps), (cap) => (
-    !cap.includes(':') && !isStandardCap(cap)
-  ));
 
   // Strip out the 'appium:' prefix
   for (let prefixedCap of prefixedCaps) {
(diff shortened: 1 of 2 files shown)


lib/basedriver/capabilities.js
```diff
@@ -101,11 +98,20 @@ function stripAppiumPrefixes (caps) {
...
-  }
+/**
+ * Get an array of all the unprefixed caps that are being used in 'alwaysMatch' and all of the 'firstMatch' object
+ * @param {Object} caps A capabilities object
+ */
+function findNonPrefixedCaps ({alwaysMatch={}, firstMatch=[]}) {
+  return _.chain([alwaysMatch, ...firstMatch])
+    .reduce((unprefixedCaps, caps) => [
+      ...unprefixedCaps,
+      ..._(caps).keys().filter((cap) => !cap.includes(':') && !isStandardCap(cap)),
```
mykola-mokhnach: what `_(caps)` is doing?

mykola-mokhnach: will it work if cap == null?

imurchie: https://lodash.com/docs/4.17.4#lodash

dpgraham: @mykola-mokhnach _(caps) allows you to chain together lodash functions to avoid callback messes. And I don't think `cap` could be null because `.keys()` returns a list of property names.

PR #175, opened 2018-01-19
Rename ELEMENT key to W3C format

dpgraham:
* If server encounters a response that has the wrong key name (if it's a W3C session and ELEMENT is provided when it should be 'element-6066-11e4-a52e-4f735466cecf'), rename all instances of that key
* Added a helper method `renameKey` that recursively finds instances of a 'key' and renames it
* Added unit tests for `renameKey`
* Added e2e test that mocks `findElements`, calls `/elements` endpoint and verifies that the payload was correctly renamed

### DIFF
--- a/lib/basedriver/helpers.js
+++ b/lib/basedriver/helpers.js
@@ -306,21 +306,21 @@ function getSwipeTouchDuration (waitGesture) {
 
 /**
  * Recursively find all instances of the key 'inKey' and rename them 'outKey'
- * @param {*} inp Any type of input
+ * @param {*} input Any type of input
  * @param {String} inKey The key name to replace
  * @param {String} outKey The key name to replace it with
  */
-function renameKey (inp, inKey, outKey) {
-  if (_.isArray(inp)) {
-    return inp.map((item) => renameKey(item, inKey, outKey));
-  } else if (_.isObject(inp)) {
-    return _.reduce(inp, (resultObj, value, key) => ({
+function renameKey (input, inKey, outKey) {
+  if (_.isArray(input)) {
+    return input.map((item) => renameKey(item, inKey, outKey));
+  } else if (_.isPlainObject(input)) {
+    return _.reduce(input, (resultObj, value, key) => ({
       ...resultObj,
       [key === inKey ? outKey : key]: renameKey(value, inKey, outKey),
     }), {});
   }
 
-  return inp;
+  return input;
 }
 
 export default { configureApp, downloadApp, downloadFile, copyLocalZip,
--- a/lib/basedriver/helpers.js
+++ b/lib/basedriver/helpers.js
@@ -318,9 +318,9 @@ function renameKey (inp, inKey, outKey) {
       ...resultObj,
       [key === inKey ? outKey : key]: renameKey(value, inKey, outKey),
     }), {});
-  } else {
-    return inp;
   }
+
+  return inp;
 }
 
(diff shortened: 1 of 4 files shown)


test/mjsonwp/mjsonwp-e2e-specs.js
```diff
@@ -456,7 +456,45 @@ describe('MJSONWP', async function () {
           delete driver.performActions;
         });
 
-        it(`should fail with a 408 error if it throws a TimeoutError exception`, async function () {
+        it(`should translate element format from MJSONWP to W3C`, async function () {
+          const W3C_ELEMENT_KEY = 'element-6066-11e4-a52e-4f735466cecf';
```
mykola-mokhnach: I'd rather move such constants on a global level

lib/basedriver/helpers.js
```diff
@@ -303,6 +303,26 @@ function getSwipeTouchDuration (waitGesture) {
...
+ */
+function renameKey (inp, inKey, outKey) {
+  if (_.isArray(inp)) {
+    return inp.map((item) => renameKey(item, inKey, outKey));
+  } else if (_.isObject(inp)) {
+    return _.reduce(inp, (resultObj, value, key) => ({
+      ...resultObj,
+      [key === inKey ? outKey : key]: renameKey(value, inKey, outKey),
+    }), {});
+  } else {
```
mykola-mokhnach: else is not needed

lib/basedriver/helpers.js
```diff
@@ -303,6 +303,26 @@ function getSwipeTouchDuration (waitGesture) {
...
+/**
+ * Recursively find all instances of the key 'inKey' and rename them 'outKey'
+ * @param {*} inp Any type of input
+ * @param {String} inKey The key name to replace
+ * @param {String} outKey The key name to replace it with
+ */
+function renameKey (inp, inKey, outKey) {
+  if (_.isArray(inp)) {
+    return inp.map((item) => renameKey(item, inKey, outKey));
+  } else if (_.isObject(inp)) {
```
mykola-mokhnach: do you mean isPlainObject?

jlipps: yeah, probably. since we're dealing with a response, we should never have something which is not a plain object. maybe we should throw in that case since it'd be a programming error on the appium side.

lib/basedriver/helpers.js
```diff
@@ -303,6 +303,26 @@ function getSwipeTouchDuration (waitGesture) {
...
 }
 
+
+/**
+ * Recursively find all instances of the key 'inKey' and rename them 'outKey'
+ * @param {*} inp Any type of input
+ * @param {String} inKey The key name to replace
+ * @param {String} outKey The key name to replace it with
+ */
+function renameKey (inp, inKey, outKey) {
```
jlipps: not sure if we get much by shortening `input` to the less-descriptive `inp`

test/basedriver/helpers-specs.js
```diff
@@ -45,4 +46,27 @@ describe('helpers', function () {
       forceWindows.restore();
     });
   });
+
+  describe('#renameKey', function () {
```
jlipps: would be good to also have a kitchen-sink test that mixes deeply-embedded keys and primitives in various fashions

dpgraham:
Changes pushed.

dpgraham:
Changes pushed again.

PR #176, opened 2018-01-19
Add W3C Error Response

dpgraham:
* Refactored `getResponseForW3CError` to return proper response 
  * Changed method so that it returns `{value: {stacktrace: '...', message: '...', error: '...'}}` (a la https://github.com/jlipps/simple-wd-spec#error-handling)
    * `message` is the same as the JSONWP error message
    * `error` is a static message indicating the type of error
    * `stacktrace` is the callstack provided in the error
  * Added an attribute called `error` to all of the MJSONWPErrors in `errors.js` which contains the static description of the W3C error
  * Also made `getResponseForW3CError` handle BadParametersError
    * `BadParametersError` isn't an `MJSONWPError` so it needs to be handled differently
* Refactored `getResponseForW3CError` and `getResponseForJsonwpError` into errors.js file

(connected to https://github.com/appium/appium/issues/10042)

### DIFF
--- a/lib/mjsonwp/errors.js
+++ b/lib/mjsonwp/errors.js
@@ -658,4 +658,31 @@ function getResponseForW3CError (err) {
   return [httpStatus, httpResBody];
 }
 
-export { MJSONWPError, errors, isErrorType, errorFromCode, getResponseForW3CError };
+function getResponseForJsonwpError (err) {
+  let httpStatus = 500;
+  let httpResBody = {
+    status: err.jsonwpCode,
+    value: {
+      message: err.message
+    }
+  };
+
+  if (isErrorType(err, errors.BadParametersError)) {
+    // respond with a 400 if we have bad parameters
+    log.debug(`Bad parameters: ${err}`);
+    httpStatus = 400;
+    httpResBody = err.message;
+  } else if (isErrorType(err, errors.NotYetImplementedError) ||
+             isErrorType(err, errors.NotImplementedError)) {
+    // respond with a 501 if the method is not implemented
+    httpStatus = 501;
+  } else if (isErrorType(err, errors.NoSuchDriverError)) {
+    // respond with a 404 if there is no driver for the session
+    httpStatus = 404;
+  }
+
+
+  return [httpStatus, httpResBody];
+}
+
+export { MJSONWPError, errors, isErrorType, errorFromCode, getResponseForW3CError, getResponseForJsonwpError };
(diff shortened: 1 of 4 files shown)


lib/mjsonwp/errors.js
```diff
@@ -1,17 +1,24 @@
 import ES6Error from 'es6-error';
 import _ from 'lodash';
-import { util } from 'appium-support';
+import { util, logger } from 'appium-support';
 import HTTPStatusCodes from 'http-status-codes';
 
+const log = logger.getLogger('MJSONWP');
+
 // base error class for all of our errors
 class MJSONWPError extends ES6Error {
```
mykola-mokhnach: at some point of time we'll need to properly rename all this stuff

lib/mjsonwp/errors.js
```diff
@@ -1,17 +1,24 @@
...
-  constructor (msg, jsonwpCode, w3cStatus = 400) {
+  constructor (msg, jsonwpCode, w3cStatus, error) {
     super(msg);
     this.jsonwpCode = jsonwpCode;
+    this.error = error || 'An unknown error has occurred';
     if (this.jsonwpCode === null) {
       this.jsonwpCode = 13;
     }
-    this.w3cStatus = w3cStatus;
+    this.w3cStatus = w3cStatus || 400;
```
mykola-mokhnach: can we have constants for these magic numbers?

lib/mjsonwp/errors.js
```diff
@@ -1,17 +1,24 @@
...
+  }
+
+  static get error () {
```
mykola-mokhnach: what is the purpose of this method?

lib/mjsonwp/errors.js
dpgraham: Yeah I should make an issue for that. Classes and files will need to be renamed.

lib/mjsonwp/errors.js
```diff
@@ -299,9 +398,12 @@ class InvalidSelectorError extends MJSONWPError {
   static code () {
     return 32;
   }
+  static error () {
```
mykola-mokhnach: I assume there should be at least one line space between class methods

lib/mjsonwp/errors.js
dpgraham: Does nothing. Will take it out.

lib/mjsonwp/errors.js
dpgraham: https://github.com/appium/appium/issues/10041

dpgraham:
Published to 2.20.2

PR #177, opened 2018-01-26
Renamed MJSONWPError to ProtocolError

dpgraham:
* Changed the logger to be 'W3C' when it's W3C protocol and 'MJSONWP' when it's MJSONWP protocol
* Also fixed an outdated test

(Note: I'll have at least 1 or 2 more PR's like this that rename MJSONWP to Protocol, I wanted to do it one-at-a-time to keep each PR simple)

### DIFF
--- a/test/jsonwp-proxy/proxy-specs.js
+++ b/test/jsonwp-proxy/proxy-specs.js
@@ -38,8 +38,8 @@ function mockProxy (opts = {}) {
 
 describe('proxy', function () {
   it('should override default params', function () {
-    let j = mockProxy({server: '127.0.0.2'});
-    j.server.should.equal('127.0.0.2');
+    let j = mockProxy({host: '127.0.0.2'});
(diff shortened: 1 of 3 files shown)


test/jsonwp-proxy/proxy-specs.js
```diff
@@ -38,8 +38,8 @@ function mockProxy (opts = {}) {
 
 describe('proxy', function () {
   it('should override default params', function () {
-    let j = mockProxy({server: '127.0.0.2'});
-    j.server.should.equal('127.0.0.2');
+    let j = mockProxy({host: '127.0.0.2'});
```
imurchie: This is the wrong direction. `server` is the right one (https://github.com/appium/appium-base-driver/blob/fa8cfaafc3930c5f1fb5d14209f2f121563183ff/lib/jsonwp-proxy/proxy.js#L17)

PR #182, opened 2018-02-01
Fixed W3C error messaging

dpgraham:
* Rename '.error' property to correct W3C property
    * The error property in the ProtocolError objects should match up with https://www.w3.org/TR/webdriver/#handling-errors under the 'JSON Error Code' column
    * Did this and added tests for all of these
* Add proxying for W3C responses 
    * If a W3C Response is provided by a proxy, map to the correct error
    * Checks if `value` is an object and `value.error` and maps it based on that (`value.error` is the W3C error code)
    * Broke off `errorFromCode` into `errorFromMJSONWPStatusCode` and `errorFromW3CJsonCode`

### DIFF
--- a/lib/protocol/errors.js
+++ b/lib/protocol/errors.js
@@ -539,8 +539,14 @@ class BadParametersError extends ES6Error {
  * for proxy failure to generate the client response.
  */
 class ProxyRequestError extends ES6Error {
-  constructor (err, jsonwp) {
-    let message = `Proxy request unsuccessful. ${util.hasValue(jsonwp) ? jsonwp.value : ''}`;
+  constructor (err, jsonwp, httpStatus, w3cError) {
+    let origMessage = '';
+    if (util.hasValue(jsonwp)) {
+      origMessage = jsonwp.value;
+    } else if (_.isPlainObject(w3cError) && w3cError.message) {
+      origMessage = w3cError.value.message;
+    }
+    let message = `Proxy request unsuccessful. ${origMessage}`;
     super(err || message);
     this.w3cStatus = HTTPStatusCodes.BAD_REQUEST;
     if (typeof jsonwp === 'string') {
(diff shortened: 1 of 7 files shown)


lib/protocol/errors.js
```diff
@@ -533,23 +539,38 @@ class BadParametersError extends ES6Error {
...
+        origMessage = responseError.value.message;
+      }
+    }
+    let message = `Proxy request unsuccessful. ${origMessage}`;
     super(err || message);
+
     this.w3cStatus = HTTPStatusCodes.BAD_REQUEST;
-    if (typeof jsonwp === 'string') {
-      this.jsonwp = JSON.parse(jsonwp);
+    if (typeof responseError === 'string') {
```
mykola-mokhnach: _.isString

lib/protocol/errors.js
```diff
@@ -613,17 +643,36 @@ function isErrorType (err, type) {
     let hasJsonwpObj = !!err.jsonwp;
```
mykola-mokhnach: this boolean conversion looks redundant to me

lib/protocol/errors.js
```diff
@@ -613,17 +643,36 @@ function isErrorType (err, type) {
...
 }
 
-// retrieve an error with the code and message
-function errorFromCode (code, message) {
-  if (code !== 13 && errorCodeMap[code]) {
-    return new errorCodeMap[code](message);
+/**
+ * Retrieve an error derived from MJSONWP status
+ * @param {number} code JSONWP status code
+ * @param {string} message
```
mykola-mokhnach: `@return` description is missing

lib/protocol/errors.js
```diff
@@ -613,17 +643,36 @@ function isErrorType (err, type) {
...
+ */
+function errorFromMJSONWPStatusCode (code, message) {
+  if (code !== 13 && jsonwpErrorCodeMap[code]) {
```
mykola-mokhnach: do we have a constant for magic number 13?

lib/protocol/errors.js
```diff
@@ -613,17 +643,36 @@ function isErrorType (err, type) {
...
+    return new jsonwpErrorCodeMap[code](message);
+  }
+  return new UnknownError(message);
+}
+
+/**
+ * Retrieve an error derived from W3C JSON Code
+ * @param {string} code W3C error string (see https://www.w3.org/TR/webdriver/#handling-errors `JSON Error Code` column)
```
mykola-mokhnach: missing message parameter description and returned value description

lib/protocol/protocol.js
```diff
@@ -295,8 +295,12 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
       }
 
       // if the status is not 0,  throw the appropriate error for status code.
-      if (util.hasValue(driverRes) && util.hasValue(driverRes.status) && parseInt(driverRes.status, 10) !== 0) {
-        throw errorFromCode(driverRes.status, driverRes.value);
+      if (util.hasValue(driverRes)) {
+        if (util.hasValue(driverRes.status) && parseInt(driverRes.status, 10) !== 0) {
```
mykola-mokhnach: I'd also check if `!isNaN(driverRes.status)` before casting it to integer

dpgraham:
Fixes pushed

lib/protocol/protocol.js
```diff
@@ -296,7 +296,7 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
 
       // if the status is not 0,  throw the appropriate error for status code.
       if (util.hasValue(driverRes)) {
-        if (util.hasValue(driverRes.status) && parseInt(driverRes.status, 10) !== 0) {
+        if (util.hasValue(driverRes.status) && parseInt(driverRes.status, 10) !== 0 && !_.isNaN(driverRes.status)) {
```
mykola-mokhnach: Use built-in javascript isNaN function rather than lodash one. Also, this check should be placed before `parseInt`

PR #184, opened 2018-02-06
Parse strings for W3C and JSON

dpgraham:
* Previously if the proxy error was a string it assumed it was jsonwp and parsed it
* Now it parses the responseError and then does it's normal business
* Added tests to `errors-specs.js` that have the fix (the `W3C --> should parse a JSON string` test failed before the fxi was put in)

### DIFF
--- a/lib/protocol/errors.js
+++ b/lib/protocol/errors.js
@@ -552,15 +552,18 @@ class ProxyRequestError extends ES6Error {
     super(err || message);
 
     this.w3cStatus = HTTPStatusCodes.BAD_REQUEST;
+
+    // If a string was provided, parse the string
     if (_.isString(responseError)) {
-      this.jsonwp = JSON.parse(responseError);
+      responseError = JSON.parse(responseError);
+    }
+
+    // If the response error is an object and value is an object, it's a W3C error (for JSONWP value is a string)
+    if (_.isPlainObject(responseError) && _.isPlainObject(responseError.value)) {
+      this.w3c = responseError.value;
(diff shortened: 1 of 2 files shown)


lib/protocol/errors.js
```diff
@@ -552,15 +552,18 @@ class ProxyRequestError extends ES6Error {
     super(err || message);
 
     this.w3cStatus = HTTPStatusCodes.BAD_REQUEST;
+
+    // If a string was provided, parse the string
     if (_.isString(responseError)) {
-      this.jsonwp = JSON.parse(responseError);
+      responseError = JSON.parse(responseError);
```
mykola-mokhnach: shouldn't this be enclosed in try/catch ?

dpgraham: I think it's actually supposed to throws an exception if the `responseError` is malformed, that way it will be caught and returned as an `UnknownError` and show the stacktrace and the JS error message.

mykola-mokhnach: ok if you say so. Is such case covered by a test?

dpgraham: I don't think so. I could add one.

dpgraham: Looks like it's impossible for `responseError` to be a non-object

The only place ProxyRequestError constructor is called is here:

https://github.com/appium/appium-base-driver/blob/master/lib/jsonwp-proxy/proxy.js#L144

...which does a JSON.parse on strings beforehand

I'd be fine with just taking the `JSON.parse` block out altogether, but I'm nervous it might break something.

PR #185, opened 2018-02-07
Change the request type of queryAppState to POST

mykola-mokhnach:
Unfortunately java client does not support sending parameters in get requests, so I must change it to POST

### DIFF
--- a/lib/protocol/routes.js
+++ b/lib/protocol/routes.js
@@ -439,6 +439,11 @@ const METHOD_MAP = {
     }
   },
   '/wd/hub/session/:sessionId/appium/device/app_state': {
+    GET: {command: 'queryAppState',
+      payloadParams: {
+        required: [['appId'], ['bundleId']]
+      }
+    },
     POST: {command: 'queryAppState',
       payloadParams: {
         required: [['appId'], ['bundleId']]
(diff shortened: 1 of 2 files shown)


lib/protocol/routes.js
```diff
@@ -439,7 +439,7 @@ const METHOD_MAP = {
     }
   },
   '/wd/hub/session/:sessionId/appium/device/app_state': {
-    GET: {command: 'queryAppState',
+    POST: {command: 'queryAppState',
```
dpgraham: Wouldn't this be a breaking change for all of the other clients?

I would rather do both:

```javascript
'/wd/hub/session/:sessionId/appium/device/app_state': {
  GET: {command: 'queryAppState', ...},
  POST: {command: 'queryAppState', ...},
```

mykola-mokhnach: it will be, but the endpoint has been added just recently, so it is not going to do any harm. Do you think we should keep the GET one?

dpgraham: If no other client uses it then I'm fine with just tossing the `GET` line.

dpgraham: Looks like the python client uses it:

https://github.com/appium/python-client/blob/f8288d5ab975c8ccfad83d9fd7d631ef987175ea/appium/webdriver/webdriver.py

mykola-mokhnach:
@dpgraham Now GET is preserved

PR #190, opened 2018-02-22
Fix the default value of payload parameters

rhysd:
Default value of payload parameters is an array(`[]`). However, as long as I saw the implementation of `makeArgs`, it seems to expect an object. So I think `{}` is correct for the default value. `[]` is working because an array is also an object in JavaScript, tho.


imurchie:
Thanks for this!

imurchie:
Published in `appium-base-driver@2.22.3`.

PR #191, opened 2018-02-23
Add logging commands to the base driver

imurchie:
Implement https://github.com/appium/appium-ios-driver/pull/267 and https://github.com/appium/appium-android-driver/pull/338/files in the base driver. The subclasses ought to just remove their implementations of `getLogTypes` and `getLog`, and make sure there is a `supportedLogTypes` property with the correct log types for that instance.

### DIFF
--- a/lib/basedriver/commands/log.js
+++ b/lib/basedriver/commands/log.js
@@ -1,0 +1,39 @@
+import log from '../logger';
+import _ from 'lodash';
+
+
+const commands = {}, helpers = {}, extensions = {};
+
+// override in sub-classes, with appropriate logs
+// in the form of
+//   {
+//     type: {
+//       description: 'some useful text',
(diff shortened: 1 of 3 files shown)


lib/basedriver/commands/log.js
```diff
@@ -0,0 +1,39 @@
...
+
+// override in sub-classes, with appropriate logs
+// in the form of
+//   {
+//     type: {
+//       description: 'some useful text',
+//       getter: () => {}, // some function that will be called to get the logs
+//     }
+//   }
+extensions.supportedLogTypes = [];
```
mykola-mokhnach: I assume it has be an object rather that an array

imurchie: Ah, yes. Thanks! This works for an empty array, since it is just an object same as `{}`.

PR #192, opened 2018-03-02
Add missing error codes

mykola-mokhnach:
This should fix https://github.com/appium/appium/issues/10286, since selenium clients such as java one decide which exception should be thrown based on the standard codes/descriptions.

### DIFF
--- a/lib/protocol/errors.js
+++ b/lib/protocol/errors.js
@@ -20,6 +20,7 @@ class ProtocolError extends ES6Error {
 }
 
 // https://github.com/SeleniumHQ/selenium/blob/176b4a9e3082ac1926f2a436eb346760c37a5998/java/client/src/org/openqa/selenium/remote/ErrorCodes.java#L215
+// https://www.w3.org/TR/webdriver/#handling-errors
 
 class NoSuchDriverError extends ProtocolError {
   static code () {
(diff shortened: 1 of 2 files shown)


dpgraham:
* Made some comments. A bunch of the errors are 404 that should be 400 (see https://www.w3.org/TR/webdriver/ for reference)
* You don't need to add the `w3cStatus()` method for 400, 400 is the default.
* Some of these commands aren't W3C commands, are you basically just taking old MJSONWP commands and making them compatible with W3C (which I'm okay with)?

mykola-mokhnach:
> Made some comments. A bunch of the errors are 404 that should be 400 (see https://www.w3.org/TR/webdriver/ for reference)

It seems like there is also a bug in Selenium implementation. I'll raise a bug to them.

> You don't need to add the w3cStatus() method for 400, 400 is the default.

I'd rather keep it (`explicit is better than implicit`).

> Some of these commands aren't W3C commands, are you basically just taking old MJSONWP commands and making them compatible with W3C (which I'm okay with)?

Java client parses the stuff in the same way, so the trick should work

mykola-mokhnach:
I merge it for now, but we need to monitor https://github.com/SeleniumHQ/selenium/issues/5562, since there might an issue in the W3C spec itself.

mykola-mokhnach:
@dpgraham Can you please publish base driver, so users can try the fix?

PR #194, opened 2018-03-14
Fix: Determine protocol in createSession

dpgraham:
* Previously, we were only checking the HTTP payload to see what protocol should be
* This is inadequate, in the case of drivers that don't proxy W3C Capabilities yet
* The fix now is to set the protocol in `createSession` to see if the W3C caps were proxied through, if they're not, set protocol to MJSONWP
* Added unit test and one e2e test to confirm that this works

(related to: https://github.com/appium/appium/issues/10350)

### DIFF
--- a/lib/basedriver/commands/session.js
+++ b/lib/basedriver/commands/session.js
@@ -17,6 +17,7 @@ commands.createSession = async function (jsonwpDesiredCapabilities, jsonwpRequir
   // Determine weather we should use jsonwpDesiredCapabilities or w3cCapabilities to get caps from
   let caps;
   if (w3cCapabilities) {
+    this.setProtocolW3C();
     if (jsonwpDesiredCapabilities) {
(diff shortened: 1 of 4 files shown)


lib/basedriver/commands/session.js
```diff
@@ -17,21 +17,26 @@ commands.createSession = async function (jsonwpDesiredCapabilities, jsonwpRequir
   // Determine weather we should use jsonwpDesiredCapabilities or w3cCapabilities to get caps from
   let caps;
   if (w3cCapabilities) {
+    this.setProtocolW3C();
```
jlipps: I guess I would have expected these to be like `this.setProtocol(BaseDriver.DRIVER_PROTOCOL.W3C)` but it's all the same

dpgraham:
@jlipps Yeah it fixes the issue with You.I driver where it wasn't ready for MJSONWP.

PR #195, opened 2018-03-16
Parse protocol from executeCommand

dpgraham:
(look at this PR first: https://github.com/appium/appium/pull/10362)

* If `executeCommand` returns an object that has a property 'protocol' and either a 'value' property or an 'error' property, adjust the response so that we respond with the protocol that the `executeCommand` function told us to use for the response
* Also: Added to delay in the 'unexpected exits' test because I was getting a race condition

### DIFF
--- a/lib/protocol/protocol.js
+++ b/lib/protocol/protocol.js
@@ -5,6 +5,7 @@ import { errors, isErrorType, ProtocolError, errorFromMJSONWPStatusCode, errorFr
 import { METHOD_MAP, NO_SESSION_ID_COMMANDS } from './routes';
 import { renameKey } from '../basedriver/helpers';
 import B from 'bluebird';
+import BaseDriver from '../basedriver/driver';
 
 
 const mjsonwpLog = logger.getLogger('MJSONWP');
@@ -198,6 +199,8 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
     let httpResBody = {};
     let httpStatus = 200;
     let newSessionId;
+    let isW3C;
+    let isMJSONWP;
 
     // get the appropriate logger depending on the protocol that is being used
     const log = getLogByProtocol(driver);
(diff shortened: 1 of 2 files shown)


lib/protocol/protocol.js
```diff
@@ -259,12 +262,29 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
...
       }
 
+      // Get the protocol after executeCommand (when command is `createSession`, protocol is assigned within
+      // createSession function)
+      isW3C = driver.isW3CProtocol();
+      isMJSONWP = driver.isMjsonwpProtocol();
+
+      // If `executeCommand` was overridden and the method returns an object
+      // with a protocol and value/error property, re-assign the protocol
+      if (_.isObject(driverRes) && _.isString(driverRes.protocol) && (!_.isNil(driverRes.value) || _.isError(driverRes.error))) {
```
imurchie: From a software engineering standpoint, I'm not a big fan of this class knowing all this info about `BaseDriver`. I'd rather `BaseDriver#executeCommand` add the appropriate info to its result, so we can just check the protocol from it.

dpgraham: That's a good idea. You're thinking that `executeCommand` should _always_ return an object of the form:

```javascript
{
  protocol: <W3C | MJSONWP>,
  value: <response value>,
  error: <response error>,
}
```

That would be do-able because afaik `appium/appium` is the only method that extends `executeCommand` and thus we wouldn't break anything else.

jlipps: I see usages of `executeCommand` in base-driver, appium itself, ios-driver, and xcuitest-driver

dpgraham: Yeah... what if we made a new method called something like `executeCommandWithProtocol` that returns the protocol and value and base driver looks for that method first 

I don't think there is a really clean solution that doesn't involve heavy rewriting.

dpgraham:
@jlipps @imurchie What do you guys think? Should we stick with this solution?

lib/protocol/protocol.js
imurchie: Adding another method that would be called in preference would not work, since the subclasses that extend `executeCommand` still need that method to be called (in ios and xcuitest cases, for async script operation).

imurchie:
I say go with this solution, so we get a fix in. It's more important, at this point, to be able to support it.

PR #196, opened 2018-03-19
Add common websocket handlers

mykola-mokhnach:
These helpers contain the common stuff, which is needed for web sockets usage. See https://github.com/appium/appium-android-driver/pull/343 for more details


lib/express/websocket.js
```diff
@@ -0,0 +1,66 @@
+import _ from 'lodash';
+import url from 'url';
+
+
+const WEBSOCKET_SERVERS = {};
+let IS_UPGRADE_LISTENER_ASSIGNED = false;
+const PATHNAME_PREFIX = '/wd/ws';
```
jlipps: given that the `/wd/hub` prefix is deprecated in w3c, maybe we should just bite the bullet and call this `/ws`?

lib/express/websocket.js
```diff
@@ -0,0 +1,66 @@
+import _ from 'lodash';
+import url from 'url';
+
+
+const WEBSOCKET_SERVERS = {};
```
jlipps: why do we have multiple servers? is it one per session? one per driver? can we encode that in messages instead?

mykola-mokhnach: it is a bit more tricky than it looks. Have you seen the example at https://github.com/websockets/ws/pull/885  ?

jlipps: I guess what I'm asking is, why do we need multiple websocket servers in the first place? Wouldn't just one do?

mykola-mokhnach: not really. I am considering an example where we have multiple parallel sessions running on a single server. Then we'll need a separate websocket for each of them. Also, there might be different websockets for different endpoints, for example one websocket may broadcast logs and the other one - screenrecording data.

jlipps: OK. that sounds reasonable. However if there's any reason to have just one server, I still think we could do it (by encoding session id and message type into the message itself). But if there's not really much expense in having multiple servers, then this seems OK to me.

mykola-mokhnach:
@imurchie WDYT? Can we merge this to try the approach?

test/basedriver/websockets-e2e-specs.js
```diff
@@ -0,0 +1,67 @@
...
+      });
+
+      removeWebsocketHandler(endpoint);
+      await new B((resolve, reject) => {
+        const client = new WebSocket(`ws://localhost:${PORT}/ws${endpoint}`);
+        client.on('message', (data) =>
+          reject(new Error(`No websocket messages are expected after the handler ` +
+                           `has been removed. '${data}' is received instead. `))
+        );
+        client.on('error', resolve);
```
imurchie: Why not reject on error?

mykola-mokhnach: because here I expect that websocket connection fails after the server has been removed

lib/express/websocket.js
```diff
@@ -0,0 +1,65 @@
...
+ * @param {Object} server - An instance of express HTTP server.
+ * @param {string} handlerPathname - Web socket endpoint path starting with
+ * single slash charcter. `/wd/ws` prefix is automatically added to the
+ * actual endpoint.
+ * @param {Object} handlerServer - WebSocket server instance. See
+ * https://github.com/websockets/ws/pull/885 for more details
+ * on how to configure the handler properly.
+ * @param {?string} pathnamePrefix ['/ws'] - Pathname prefix.
+ */
+function addWebsocketHandler (server, handlerPathname, handlerServer, pathnamePrefix = PATHNAME_PREFIX) {
```
imurchie: Isn't `WebSocket` with a capital `S`? It would be good to reflect that in the function names.

mykola-mokhnach: it is not a problem at all to change that

mykola-mokhnach:
> How is this intended to be used? If we have the instance of the express server already, could this not be added to the server, so we could just go await server.addWebSocketHandler(...)?

I like the idea, would you mind to provide some example on how to properly attach these methods to the existing server instance?

imurchie:
The express server is always created (whether Appium's, or one of the driver running separately) using https://github.com/appium/appium-base-driver/blob/master/lib/express/server.js#L16-L64

So, you could just add the functions to the instance of the server that we create
```js
import { addWebSocketHandler, removeWebSocketHandler } from '.websocket';

let app = express();
app.addWebSocketHandler = addWebSocketHandler;
app.removeWebSocketHandler = removeWebSocketHandler;
```

mykola-mokhnach:
Will it set `this` context properly for them or I need to use `bind`?

imurchie:
If called on the server instance, `this` will be set correctly. `this` is determined by call site.

https://github.com/getify/You-Dont-Know-JS/blob/master/this%20%26%20object%20prototypes/ch2.md

mykola-mokhnach:
Yep, thanks. Excuse my stupidity %)

mykola-mokhnach:
How do you like it now @imurchie ?

imurchie:
Published in `appium-base-driver@2.25.0`.

jlipps:
🎉👌🏼

PR #199, opened 2018-03-24
Add endpoints for system clipboard interaction

mykola-mokhnach:
I would like to generalize iOS and Android clipboards interaction primitives and create high-level wrappers over them

### DIFF
--- a/lib/protocol/routes.js
+++ b/lib/protocol/routes.js
@@ -597,6 +597,20 @@ const METHOD_MAP = {
   '/wd/hub/session/:sessionId/element/:elementId/property/:name': {
     GET: {command: 'getProperty'}
   },
+  '/wd/hub/session/:sessionId/appium/device/set_clipboard': {
+    POST: {command: 'setClipboard', payloadParams: {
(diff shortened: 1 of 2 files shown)


lib/protocol/routes.js
```diff
@@ -597,6 +597,22 @@ const METHOD_MAP = {
...
+    POST: {command: 'setClipboard', payloadParams: {
+      required: ['content'],
+      optional: [
+        'contentType',
+        'label',
+      ]},
+    }
+  },
+  '/wd/hub/session/:sessionId/appium/device/get_clipboard': {
+    POST: {command: 'getClipboard', payloadParams: {
```
imurchie: This being a `POST` request seems weird. Is the `contentType` necessary?

mykola-mokhnach: this is optional, but I'd like to keep it, so one can filter clipboard content. Also, this could be extended later to get the content of multiple clipboard items and not only the primary one

imurchie: Alright.

PR #201, opened 2018-04-03
Fix protocol parsing

dpgraham:
* Some problems with parsing `{protocol: <protocol>, value: <value>}`
  * `(!_.isNil(driverRes.value)` was incorrect check because `value` can be undefined or null; replaced it with `_.has(driverRes, 'value')` to check that the `value` property was set
  *  ```
       if (driverRes.value) {
          driverRes = driverRes.value;
        } 
      ``` 

      was also incorrect because `driverRes.value` may be false-y

* Moved logic into it's own function and then added tests to cover the problem cases

### DIFF
--- a/lib/protocol/protocol.js
+++ b/lib/protocol/protocol.js
@@ -450,7 +450,7 @@ async function doJwpProxy (driver, req, res) {
 
 /**
  * Check a driver command respond and see if the protocol and value was passed in
- * @param {Object} driverRes
+ * @param {Object} driverRes Response returned by `executeCommand` in an inner driver
  * @throws Will throw an error if driverRes.error is provided
  */
 function parseProtocol (driverRes) {
--- a/lib/protocol/protocol.js
+++ b/lib/protocol/protocol.js
@@ -450,7 +450,7 @@ async function doJwpProxy (driver, req, res) {
 
 /**
  * Check a driver command respond and see if the protocol and value was passed in
- * @param {Object} driverRes Response returned by `executeCommand` in an inner driver
+ * @param {Object} driverRes Response returned by `executeCommand` in an inner driver 
  * @throws Will throw an error if driverRes.error is provided
  */
 function parseProtocol (driverRes) {
--- a/lib/protocol/protocol.js
+++ b/lib/protocol/protocol.js
@@ -453,14 +453,18 @@ async function doJwpProxy (driver, req, res) {
  * @param {Object} driverRes
  */
 function parseProtocol (driverRes) {
-  if (_.isObject(driverRes) && _.isString(driverRes.protocol)) {
-    if (_.has(driverRes, 'value')) {
-      let isW3C = driverRes.protocol === BaseDriver.DRIVER_PROTOCOL.W3C;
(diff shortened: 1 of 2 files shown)


lib/protocol/protocol.js
```diff
@@ -451,5 +448,21 @@ async function doJwpProxy (driver, req, res) {
   }
 }
 
+/**
+ * Check a driver command respond and see if the protocol and value was passed in
+ * @param {Object} driverRes
```
mykola-mokhnach: missing `return` description and `driverRes`

lib/protocol/protocol.js
```diff
@@ -451,5 +448,21 @@ async function doJwpProxy (driver, req, res) {
...
+ */
+function parseProtocol (driverRes) {
+  if (_.isObject(driverRes) && _.isString(driverRes.protocol)) {
```
mykola-mokhnach: isPlainObject?

mykola-mokhnach: I'd rather replace it with 

```
if (!_.isPlainObject(driverRes) || !_.isString(driverRes.protocol)) {
  return null;
}
```

lib/protocol/protocol.js
mykola-mokhnach: this function should also have `@throws` description

lib/protocol/protocol.js
```diff
@@ -451,5 +448,21 @@ async function doJwpProxy (driver, req, res) {
...
+    if (_.has(driverRes, 'value')) {
+      let isW3C = driverRes.protocol === BaseDriver.DRIVER_PROTOCOL.W3C;
+      let isMJSONWP = driverRes.protocol === BaseDriver.DRIVER_PROTOCOL.MJSONWP;
+      return {isW3C, isMJSONWP, value: driverRes.value};
+    } else if (_.isError(driverRes.error)) {
```
mykola-mokhnach: else if redundant

dpgraham:
Fixes pushed

lib/protocol/protocol.js
```diff
@@ -451,5 +448,26 @@ async function doJwpProxy (driver, req, res) {
   }
 }
 
+/**
+ * Check a driver command respond and see if the protocol and value was passed in
+ * @param {Object} driverRes Response returned by `executeCommand` in an inner driver
+ * @throws Will throw an error if driverRes.error is provided
```
mykola-mokhnach: how about returns?

lib/protocol/protocol.js
```diff
@@ -451,5 +448,26 @@ async function doJwpProxy (driver, req, res) {
...
+
+  if (_.has(driverRes, 'value')) {
+    let isW3C = driverRes.protocol === BaseDriver.DRIVER_PROTOCOL.W3C;
+    let isMJSONWP = driverRes.protocol === BaseDriver.DRIVER_PROTOCOL.MJSONWP;
+    return {isW3C, isMJSONWP, value: driverRes.value};
+  }
+
+  if (_.isError(driverRes.error)) {
+    throw driverRes.error;
+  }
```
mykola-mokhnach: I prefer to explicitly return null, since it improves readability

dpgraham:
Was published.

PR #203, opened 2018-04-18
Add an universal endpoint for images comparison

mykola-mokhnach:
Based on https://github.com/appium/appium-support/pull/65

### DIFF
--- a/lib/basedriver/commands/images.js
+++ b/lib/basedriver/commands/images.js
@@ -8,9 +8,9 @@ const MATCH_FEATURES_MODE = 'matchFeatures';
 const GET_SIMILARITY_MODE = 'getSimilarity';
 const MATCH_TEMPLATE_MODE = 'matchTemplate';
 
-commands.compareImages = async function (mode, firstImageData, secondImageData, options = {}) {
-  const img1 = Buffer.from(firstImageData, 'base64');
-  const img2 = Buffer.from(secondImageData, 'base64');
+commands.compareImages = async function (mode, firstImage, secondImage, options = {}) {
+  const img1 = Buffer.from(firstImage, 'base64');
+  const img2 = Buffer.from(secondImage, 'base64');
   switch (_.toLower(mode)) {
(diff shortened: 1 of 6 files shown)


lib/basedriver/commands/images.js
```diff
@@ -0,0 +1,29 @@
...
+const commands = {}, helpers = {}, extensions = {};
+
+const MATCH_FEATURES_MODE = 'matchFeatures';
+const GET_SIMILARITY_MODE = 'getSimilarity';
+const MATCH_TEMPLATE_MODE = 'matchTemplate';
+
+commands.compareImages = async function (mode, firstImageData, secondImageData, options = {}) {
+  const img1 = Buffer.from(firstImageData, 'base64');
+  const img2 = Buffer.from(secondImageData, 'base64');
+  switch (_.toLower(mode)) {
```
jlipps: why use both `_.toLower` and `String.toLowerCase`?

lib/protocol/routes.js
```diff
@@ -613,6 +613,12 @@ const METHOD_MAP = {
       ]},
     }
   },
+  '/wd/hub/session/:sessionId/appium/compare_images': {
+    POST: {command: 'compareImages', payloadParams: {
+      required: ['mode', 'firstImageData', 'secondImageData'],
```
jlipps: thinking about the param names, why not just `firstImage` and `secondImage`, or `image1` and `image2`? We'll specify the data type as b64-encoded PNG, but the fact that it is data seems superfluous.

lib/basedriver/commands/images.js
```diff
@@ -0,0 +1,29 @@
+import _ from 'lodash';
+import { errors } from '../../protocol/errors';
+import { imageUtil } from 'appium-support';
```
jlipps: do we need to update package.json to ensure `appium-support` is at a version which has this library method?

mykola-mokhnach: yep, I'm waiting until the module is published

PR #210, opened 2018-05-12
Properly parse error codes for different protocols

mykola-mokhnach:
The `parseProtocol` method should properly return protocol type in the first priority and should not throw until the protocol is set properly. This fixes wrong responses in case of driver failures (like in https://github.com/appium/appium/issues/10681). The only thing I cannot figure out yet is on which stage the original error message is lost, because the resulting one contains the default text of the corresponding protocol error (upd: figured it out - the problem lies inside bootstrap module):

> [debug] [AndroidBootstrap] [BOOTSTRAP LOG] [debug] Returning result: {"status":32,"value":"javax.xml.transform.TransformerException: Expected ), but found: $"}

vs

> org.openqa.selenium.InvalidSelectorException: Argument was an invalid selector (e.g. XPath/CSS). (WARNING: The server did not provide any stacktrace information)

### DIFF
--- a/lib/protocol/protocol.js
+++ b/lib/protocol/protocol.js
@@ -463,7 +463,7 @@ function parseProtocol (driverRes) {
   if (_.has(driverRes, 'value')) {
     return {isW3C, isMJSONWP, value: driverRes.value};
   }
-  if (_.has(driverRes, 'error') && _.isPlainObject(driverRes.error)) {
+  if (_.has(driverRes, 'error')) {
     if (isMJSONWP) {
       return {isW3C, isMJSONWP, value: {
         status: driverRes.error.jsonwpCode,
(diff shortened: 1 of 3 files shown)


lib/protocol/protocol.js
```diff
@@ -451,22 +451,33 @@ async function doJwpProxy (driver, req, res) {
...
-
-  if (_.isError(driverRes.error)) {
-    throw driverRes.error;
+  if (_.has(driverRes, 'error')) {
+    if (isMJSONWP) {
+      return {isW3C, isMJSONWP, value: {
+        status: driverRes.error.jsonwpCode,
+        value: driverRes.error.message,
+      }};
+    }
```
dpgraham: You don't need to translate the error. The error should be passed along on it's own. The functions `getResponseForW3CError` and `getResponseForMJSONWPError` parses the `Error` and then responds with the status/value.

mykola-mokhnach:
Closed in favour of https://github.com/appium/appium-base-driver/pull/211

PR #211, opened 2018-05-13
Fix: Throw error after parsing protocol from proxied response

dpgraham:
* Before it was throwing immediately which meant we weren't parsing the protocol and weren't showing the proper error responses
* If proxied W3C error doesn't have an error, make it `Unknown`
* Also removed usage of `static` because it's not part of standards yet


test/protocol/protocol-specs.js
```diff
@@ -31,7 +31,9 @@ describe('Protocol', async function () {
       parseProtocol({protocol: 'MJSONWP', value}).should.eql({isW3C: false, isMJSONWP: true, value});
     });
     it('should throw if {protocol: "MJSONWP", error}', function () {
-      (() => parseProtocol({protocol: 'W3C', error: new Error('some error')})).should.throw(/some error/);
+      //(() => ).should.throw(/some error/);
```
mykola-mokhnach: do we need the comment?

lib/protocol/protocol.js
```diff
@@ -273,6 +273,9 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
       if (parsedDriverRes) {
         isW3C = parsedDriverRes.isW3C;
         isMJSONWP = parsedDriverRes.isMJSONWP;
+        if (driverRes.error) {
```
mykola-mokhnach: driverRes or parsedDriverRes?

lib/protocol/errors.js
```diff
@@ -811,6 +811,10 @@ function getResponseForW3CError (err) {
 
   httpStatus = err.w3cStatus;
 
+  if (!error) {
```
mykola-mokhnach: it might be handy to mention that `error` here is the actual string signature

lib/basedriver/driver.js
```diff
@@ -412,4 +398,18 @@ for (let [cmd, fn] of _.toPairs(commands)) {
   BaseDriver.prototype[cmd] = fn;
 }
 
+BaseDriver.DRIVER_PROTOCOL = {
```
mykola-mokhnach: static look better AFAIK

lib/protocol/protocol.js
dpgraham: They would both have the same error, but I'll change it to `parsedDriverRes` because that's more obvious.

dpgraham:
Fixes pushed

lib/protocol/errors.js
```diff
@@ -804,16 +804,20 @@ function getResponseForW3CError (err) {
   if (isErrorType(err, errors.BadParametersError)) {
```
mykola-mokhnach: why this condition is needed?

dpgraham: JSONWP are almost all 500, except for `BadParameters`, `NotYetImplemented` and `NoSuchDriver`.

lib/protocol/protocol.js
```diff
@@ -465,14 +468,15 @@ function parseProtocol (driverRes) {
     return null;
   }
 
+  let isW3C = driverRes.protocol === BaseDriver.DRIVER_PROTOCOL.W3C;
```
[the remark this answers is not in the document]
mykola-mokhnach: the function does not throw anymore. Also the returned result might be different

lib/protocol/errors.js
mykola-mokhnach: this might be handy as a comment

lib/protocol/errors.js
```diff
@@ -804,16 +804,20 @@ function getResponseForW3CError (err) {
...
+  if (!w3cErrorString) {
+    w3cErrorString = UnknownError.error();
+  }
+
   let httpResBody = {
     value: {
-      error,
+      error: w3cErrorString,
       message: err.message,
       stacktrace: err.stack,
```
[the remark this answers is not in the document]
dpgraham: This gives us the stacktrace that was provided by the inner driver. So in the case of a Java inner driver being proxied, we would get the Java stack, which IMO is more helpful.

dpgraham:
Published to 2.30.1

PR #213, opened 2018-05-27
Enforce UTF-8 encoding on request headers

mykola-mokhnach:
This should resolve issues like https://github.com/appium/ruby_lib/issues/598#issuecomment-392308793, since some HTTP servers fall back to ASCII instead of UTF-8 if the content encoding is not set explicitly in request headers.

@KazuCocoa FYI


mykola-mokhnach:
Closed in favour of https://github.com/appium/appium-base-driver/pull/214

PR #218, opened 2018-06-06
Fix sending of commands by JWProxy  class

mykola-mokhnach:
This method should throw specialised errors instead of the generic one. Also, it supports W3C responses now.

See https://github.com/appium/appium-xcuitest-driver/pull/688#issuecomment-395065421 for more details.

Also, it looks like WDA always returns returns response code as 200 independently of the actual response status.


imurchie:
Published in `appium-base-driver@2.32.0`.

PR #220, opened 2018-06-08
Avoid throwing proxy request error inside command call

mykola-mokhnach:
The second PR to resolve https://github.com/appium/appium/issues/10809#issuecomment-395636141

### DIFF
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -144,8 +144,8 @@ class JWProxy {
       try {
         responseError = JSON.parse(e.error);
       } catch (e1) {
-        if (_.isString(e.error) && e.error.length) {
-          log.warn(`Got unexpected response: ${_.truncate(e.error, {length: 300})}`);
+        if (!_.isEmpty(e.error) && _.isString(e.error)) {
+          log.warn(`Got an unexpected response: ${_.truncate(e.error, {length: 300})}`);
         }
         responseError = _.isPlainObject(e.error) ? e.error : null;
       }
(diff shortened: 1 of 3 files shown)


lib/jsonwp-proxy/proxy.js
```diff
@@ -156,9 +156,18 @@ class JWProxy {
...
+      [response, resBody] = await this.proxy(url, method, body);
+    } catch (err) {
+      if (err instanceof errors.ProxyRequestError) {
+        throw err.getActualError();
+      }
+      throw new errors.UnknownError(err.message);
+    }
     resBody = util.safeJsonParse(resBody);
-    if (util.hasValue(resBody.status)) {
+    if (_.has(resBody, 'status') && util.hasValue(resBody.status)) {
```
imurchie: I don't understand the value of adding the `_.has(resBody, 'status')` check:
```js
let resBody = {
  a: undefined,
  b: null,
  c: 'something',
};

_.has(resBody, 'a');
// true
_.has(resBody, 'b');
// true
_.has(resBody, 'c');
// true
_.has(resBody, 'd');
// false

util.hasValue(resBody.a);
// false
util.hasValue(resBody.b);
// false
util.hasValue(resBody.c);
// true
util.hasValue(resBody.c);
// false
```
So any case that is rejected by `_.has` would also be rejected by `util.hasValue`.

mykola-mokhnach: this is for the case if `resBody` is not an object. Let say if it is `null` than calling resBody.status will immediately throw an error

imurchie: Ok. I'd rather that be explicit, myself (like, check if `resBody` is an object).

mykola-mokhnach: do you have a better idea on how to do such verification?

dpgraham: You could make a conditional outside of it like

```javascript
if (_.isPlainObject(resBody)) {
  if (util.hasValue(resBody.a) {
    ...
  }
}
```

imurchie: It seems like something is wrong if resBody is not an object. At least, the code will either return the non-object resBody (if response.statusCode is 200) or throw an error. 

So there could be an initial check if it is not an object, and either return it or throw the error, then the rest can just assume it is an object. Or each of those _.has checks can be an isObject check. Or just an existence check, since any random property can be attempted to be accessed from a non-null, non-undefined variable, with no error (e.g., `a = 2; a.status;` will not throw an error).

PR #229, opened 2018-06-28
Remove list of valid automation names

imurchie:
The automation names are the property of `AppiumDriver`, not the base driver. No other drivers ought to care about them.

See https://github.com/appium/appium/pull/10967


mykola-mokhnach:
I assume we should raise the major module version after this change is merged

imurchie:
Yes, though as far as I can tell this function was only ever used in the AppiumDriver. But better safe than sorry.

PR #230, opened 2018-07-05
add mjpeg stream helper class to basedriver

jlipps:
There is this "standard" of sorts called [MJPEG](https://en.wikipedia.org/wiki/Motion_JPEG), which allows for streaming of images over HTTP. Some users of Appium have MJPEG servers connected to the devices they use with Appium, which deliver relatively high-framerate images off the device. If this is the case, why not just use it for screenshots as well and bypass the sometimes time-consuming Appium-internal screenshot commands?

MJPEG is a sort-of "push" mechanism, so in this PR I just provide a class which maintains a reference to the last-pushed image, which would then be available to any driver that wants it. I'll start the driver PRs soon, but the basic idea is that this will be behind a capability, which if turned on will just send back the last-received image immediately upon screenshot request. Something like this:

```js
// in driver session start
if (this.opts.mjpegScreenshotUrl) {
  this.mjpegStream = new MJpegStream(this.opts.mjpegScreenshotUrl);
  this.mjpegStream.start();
}
```

```js
// in screenshot command implementation 
if (this.mjpegStream && this.mjpegStream.lastChunkBase64) {
  return this.mjpegStream.lastChunkBase64;
}

// now continue with regular screenshot logic as a fallback
```

```js
// in driver session stop/cleanup
if (this.mjpegStream) {
  this.mjpegStream.stop();
}
```


lib/basedriver/mjpeg.js
```diff
@@ -0,0 +1,84 @@
...
+    this.url = mJpegUrl;
+    this.clear();
+  }
+
+  /**
+   * Get the base64-encoded version of the JPEG
+   * @returns {string}
+   */
+  get lastChunkBase64 () {
+    if (!this.lastChunk) {
```
mykola-mokhnach: Ternary operator might be shorter here

lib/basedriver/mjpeg.js
```diff
@@ -0,0 +1,84 @@
...
+    this.request = null;
+    this.consumer = null;
+    this.lastChunk = null;
+    this.updateCount = 0;
+  }
+
+  /**
+   * Start reading the MJpeg stream and storing the last image
+   */
+  start () {
```
mykola-mokhnach: why not async?

lib/basedriver/mjpeg.js
```diff
@@ -0,0 +1,84 @@
...
+
+    this.request
+      .on('error', onErr)   // ensure we do something with errors
+      .pipe(this.consumer)  // allow chunking and transforming of jpeg data
+      .pipe(this);          // send the actual jpegs to ourself
+  }
+
+  /**
+   * Stop reading the MJpeg stream. Ensure we disconnect all the pipes and stop
+   * the HTTP request itself. Then reset the state.
```
mykola-mokhnach: It might be useful to mark overridden method with `@Override`

test/basedriver/fixtures/mjpeg-server.js
```diff
@@ -0,0 +1,24 @@
...
+
+export function initMJpegServer (port, intMs = 300, times = 20) {
+  const server = http.createServer(async function (req, res) {
+    const mJpegReqHandler = mJpegServer.createReqHandler(req, res);
+    const jpg = await fs.readFile(IMAGE_PATH, 'binary');
+
+    // just send the same jpeg over and over
+    for (let i = 0; i < times; i++) {
+      await B.delay(intMs);
+      mJpegReqHandler._write(jpg, null, () => {});
```
mykola-mokhnach: there is also _.noop helper

mykola-mokhnach:
We probably would need some fast PNG->JPEG converter here as well

mykola-mokhnach:
> For now my thought is if someone uses this, they know they will get jpeg formatted screenshots out of appium instead of png.

Selenium's getScreenshot always returns images in PNG format. There are no other endpoints, which allow to get JPEG directly. That is why I suppose we need to do the conversion on Appium side

lib/basedriver/mjpeg.js
```diff
@@ -0,0 +1,105 @@
...
+    this.url = mJpegUrl;
+    this.clear();
+  }
+
+  /**
+   * Get the base64-encoded version of the JPEG
+   * @returns {string}
+   */
+  get lastChunkBase64 () {
+    return this.lastChunk ? this.lastChunk.toString('base64') : null;
```
mykola-mokhnach: _.isBuffer(lastChunk) ?

lib/basedriver/mjpeg.js
```diff
@@ -0,0 +1,105 @@
...
+    this.request = null;
+    this.consumer = null;
+    this.lastChunk = null;
+    this.updateCount = 0;
+  }
+
+  /**
+   * Start reading the MJpeg stream and storing the last image
+   */
+  async start () {
```
mykola-mokhnach: 👍

lib/basedriver/mjpeg.js
```diff
@@ -0,0 +1,105 @@
...
+
+    await startPromise;
+  }
+
+  /**
+   * Stop reading the MJpeg stream. Ensure we disconnect all the pipes and stop
+   * the HTTP request itself. Then reset the state.
+   */
+  stop () {
+    if (this.consumer) {
```
mykola-mokhnach: if (!this.consumer) return

lib/basedriver/mjpeg.js
```diff
@@ -0,0 +1,105 @@
...
+    };
+
+    this.request = request(this.url);
+
+    this.request
+      .on('error', onErr)   // ensure we do something with errors
+      .pipe(this.consumer)  // allow chunking and transforming of jpeg data
+      .pipe(this);          // send the actual jpegs to ourself
+
+    await startPromise;
```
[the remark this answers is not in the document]
jlipps: there is no last image, so that's not possible.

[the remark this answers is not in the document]
mykola-mokhnach: I mean this might block forever if the server does not produce any data

lib/basedriver/mjpeg.js
jlipps: this method is not overridden

lib/basedriver/mjpeg.js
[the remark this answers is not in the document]
jlipps: oh right. that would be an error. probably we should have a timeout and if the timeout gets hit, throw an error

lib/basedriver/mjpeg.js
```diff
@@ -0,0 +1,105 @@
...
+  get lastChunkBase64 () {
+    return this.lastChunk ? this.lastChunk.toString('base64') : null;
+  }
+
+
+  /**
+   * Reset internal state
+   */
+  clear () {
+    this.didStart = false;
```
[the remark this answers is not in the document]
jlipps: nope! will remove

jlipps:
@mykola-mokhnach @imurchie what do you think, should this be a part of appium-support instead? now that it is just a class which can be used, but not tied to the driver code itself.

mykola-mokhnach:
If the class has no direct dependencies on base driver then yes, it would be better to keep it in appium-support

lib/basedriver/mjpeg.js
```diff
@@ -0,0 +1,119 @@
...
+    // use the deferred pattern so we can wait for the start of the stream
+    // based on what comes in from an external pipe
+    let startPromise = new B((res, rej) => {
+      this.registerStartSuccess = res;
+      this.registerStartFailure = rej;
+    });
+
+    // start a timeout so that if the server does not return data, we don't
+    // block forever.
+    setTimeout(() => {
```
mykola-mokhnach: I think bluebird promises already have the standard .timeout method

jlipps:
closing, gonna move to appium-support

PR #231, opened 2018-07-06
Handle /timeouts W3C + MJSONWP proxying

dpgraham:
* The /timeouts endpoint is a special case because the syntax differs between W3C and MJSONWP
* W3C takes a body like this:

  ```javascript
  {pageLoad: 1, script: 2, implicit: 3}
  ```

* MJSONWP takes a body like this

  ```javascript
  {type: 'script', ms: 2}
  ```

* Since W3C can send > 1 timeout type, we need to have special logic so that if a W3C /timeouts request is being proxied to a MJSONWP /timeouts endpoint, we need to proxy 1 request for each timeout
* The way this PR handles that is by creating an array of timeout request objects like this:

  If the incoming request is W3C, like this:

```javascript
{pageLoad: 1, script: 2, implicit: 3}
```

It will generate an array that looks like this:

```javascript
[
  {pageLoad: 1, script: 2, implicit: 3, type: 'pageLoad', ms: 1},
  {pageLoad: 1, script: 2, implicit: 3, type: 'script', ms: 2},
  {pageLoad: 1, script: 2, implicit: 3, type: 'implicit', ms: 3},
]
```

  If the incoming request is MJSONWP, like this:

```javascript
{type: 'pageLoad', ms: 100}
```

  It will generate an array that looks like this:

```javascript
[{pageLoad: 100, script: undefined, implicit: undefined, type: 'pageLoad', ms: 100}]
```

* All of the items in the request objects array are compatible with W3C and MJSONWP
* What base driver does is executes them one-by-one
  * If it encounters a W3C proxied response, it will return that response immediately
  * If it encounters an MJSONWP proxied response, and it's successful, it will continue to the next 
  object and proxy that one too
  * If it encounters an MJSONWP proxied response, and it fails, it will respond with that failure and 
  not continue to the next

(related to: https://github.com/appium/appium/issues/11016)

### DIFF
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -163,7 +163,8 @@ class JWProxy {
    * at a time. So if we're using W3C and proxying to MJSONWP and there's more than one timeout type
    * provided in the request, we need to do 3 proxies and combine the result
    *
-   * @param {Array} timeoutObjects List of MJSONWP + W3C compatible timeouts request objects
+   * @param {Object} body Request body
+   * @return {Array} Array of W3C + MJSONWP compatible timeout objects
    */
   getTimeoutRequestObjects (body) {
     const {protocol} = body;
@@ -183,20 +184,22 @@ class JWProxy {
           });
         }
       }
-    } else {
-      timeoutObjects.push({
-        ...baseObj,
-        [body.type]: body.ms,
-        type: body.type,
-        ms: body.ms,
-      });
+      return timeoutObjects;
     }
-    return timeoutObjects;
+
+    return [{
+      ...baseObj,
+      [body.type]: body.ms,
+      type: body.type,
+      ms: body.ms,
+    }];
   }
 
   /**
    * Proxy an array of timeout objects and merge the result
-   * @param {Array} timeoutObjects List of W3C+MJSONWP compatible /timeouts requests
+   * @param {String} url Endpoint url
+   * @param {String} method Endpoint method
+   * @param {Object} body Request body
    */
   async proxySetTimeouts (url, method, body) {
     let response, resBody;
@@ -232,7 +235,7 @@ class JWProxy {
     let response;
     let resBody;
     try {
-      if (url === '/timeouts' && method.toLowerCase() === 'post') {
+      if (url.endsWith('timeouts') && method.toLowerCase() === 'post') {
         [response, resBody] = await this.proxySetTimeouts(url, method, body);
       } else {
         [response, resBody] = await this.proxy(url, method, body);
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -3,12 +3,15 @@ import { logger, util } from 'appium-support';
 import request from 'request-promise';
 import { getSummaryByCode } from '../jsonwp-status/status';
 import { errors, isErrorType, errorFromMJSONWPStatusCode, errorFromW3CJsonCode } from '../protocol/errors';
+import BaseDriver from '../basedriver/driver';
 
 
 const log = logger.getLogger('JSONWP Proxy');
 // TODO: Make this value configurable as a server side capability
 const LOG_OBJ_LENGTH = 1024; // MAX LENGTH Logged to file / console
 const DEFAULT_REQUEST_TIMEOUT = 240000;
+
+const {MJSONWP, W3C} = BaseDriver.DRIVER_PROTOCOL;
 
 class JWProxy {
   constructor (opts = {}) {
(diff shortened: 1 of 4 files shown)


lib/jsonwp-proxy/proxy.js
```diff
@@ -155,19 +158,94 @@ class JWProxy {
     return [res, resBody];
   }
 
+  /**
+   * W3C /timeouts can take as many as 3 timeout types at once, MJSONWP /timeouts only takes one
+   * at a time. So if we're using W3C and proxying to MJSONWP and there's more than one timeout type
+   * provided in the request, we need to do 3 proxies and combine the result
+   *
+   * @param {Array} timeoutObjects List of MJSONWP + W3C compatible timeouts request objects
```
mykola-mokhnach: param is `body`

mykola-mokhnach: also no return description

lib/jsonwp-proxy/proxy.js
```diff
@@ -155,19 +158,94 @@ class JWProxy {
...
+    const {protocol} = body;
+    const timeoutObjects = [];
+    const baseObj = {
+      script: body.script,
+      pageLoad: body.pageLoad,
+      implicit: body.implicit,
+    };
+    if (protocol === W3C) {
+      for (let [type, ms] of _.toPairs(baseObj)) {
+        if (_.isNumber(ms)) {
```
mykola-mokhnach: should we also check if type is defined?

lib/jsonwp-proxy/proxy.js
```diff
@@ -155,19 +158,94 @@ class JWProxy {
...
+        type: body.type,
+        ms: body.ms,
+      });
+    }
+    return timeoutObjects;
+  }
+
+  /**
+   * Proxy an array of timeout objects and merge the result
+   * @param {Array} timeoutObjects List of W3C+MJSONWP compatible /timeouts requests
```
mykola-mokhnach: same documentation issue here

lib/jsonwp-proxy/proxy.js
```diff
@@ -155,19 +158,94 @@ class JWProxy {
...
+   */
+  async proxySetTimeouts (url, method, body) {
+    let response, resBody;
+    for (let timeoutObj of this.getTimeoutRequestObjects(body)) {
+      [response, resBody] = await this.proxy(url, method, timeoutObj);
+
+      let protocol = this.getProtocolFromResBody(resBody);
+
+      // If we got a non-MJSONWP response, return the result, nothing left to do
+      if (protocol !== MJSONWP) {
```
mykola-mokhnach: I think this is redundant, since there will be anyway only one item in the array in case of mjsonwp

lib/jsonwp-proxy/proxy.js
```diff
@@ -155,19 +158,94 @@ class JWProxy {
...
+    }
+    return [response, resBody];
+  }
+
+  getProtocolFromResBody (resBody) {
+    if (_.isPlainObject(resBody) && util.hasValue(resBody.status)) {
+      return MJSONWP;
+    } else if (util.hasValue(resBody.value)) {
+      return W3C;
+    }
```
mykola-mokhnach: what if both conditions fail?

lib/jsonwp-proxy/proxy.js
```diff
@@ -155,19 +158,94 @@ class JWProxy {
...
+  }
+
   async command (url, method, body = null) {
     let response;
     let resBody;
     try {
-      [response, resBody] = await this.proxy(url, method, body);
+      if (url === '/timeouts' && method.toLowerCase() === 'post') {
```
mykola-mokhnach: I don't really like the fact that the url is hardcoded here. We might make it more flexible

lib/jsonwp-proxy/proxy.js
```diff
@@ -155,19 +158,94 @@ class JWProxy {
...
+      for (let [type, ms] of _.toPairs(baseObj)) {
+        if (_.isNumber(ms)) {
+          timeoutObjects.push({
+            ...baseObj,
+            type,
+            ms,
+          });
+        }
+      }
+    } else {
```
mykola-mokhnach:
```
return [{
             ...baseObj,
             type,
             ms,
           }];
```

mykola-mokhnach: `else` will be redundant in such case

lib/jsonwp-proxy/proxy.js
dpgraham: `type` will always be defined because it's the key of an object

lib/jsonwp-proxy/proxy.js
dpgraham: If there's an array of timeout objects like this:

```javascript
[
  {pageLoad: 1, script: 2, implicit: 3, type: 'pageLoad', ms: 1},
  {pageLoad: 1, script: 2, implicit: 3, type: 'script', ms: 2},
  {pageLoad: 1, script: 2, implicit: 3, type: 'implicit', ms: 3},
]
```

...then we need to stop after the first call if it's W3C because it's done.

lib/jsonwp-proxy/proxy.js
dpgraham: I'd like to return it as `undefined` so that it's just an unknown protocol.

dpgraham:
Fixes pushed

lib/jsonwp-proxy/proxy.js
```diff
@@ -155,19 +158,97 @@ class JWProxy {
...
+      return W3C;
+    }
+  }
+
   async command (url, method, body = null) {
     let response;
     let resBody;
     try {
-      [response, resBody] = await this.proxy(url, method, body);
+      if (url.endsWith('timeouts') && method.toLowerCase() === 'post') {
```
jlipps: maybe `endsWith('/timeouts')`? what if a user is trying to get the 'timeouts' attribute of an element? (i know that's a 'get', just saying...)

PR #235, opened 2018-07-15
add -image locator strategy and related logic

jlipps:
First stab at putting the find by image logic into base driver.

I know there's no tests and docstrings, will add those. Just wanted to get the ideas here for any initial comments.

I haven't yet done extensive testing, but the happy path works for me locally.

### DIFF
--- a/lib/basedriver/commands/find.js
+++ b/lib/basedriver/commands/find.js
@@ -1,7 +1,16 @@
 import log from '../logger';
+import Jimp from 'jimp';
+import Promise from 'bluebird';
+import { errors } from '../../..';
+import { Buffer } from 'buffer';
+import { DEFAULT_MATCH_THRESHOLD, MATCH_TEMPLATE_MODE } from './images';
+import { W3C_ELEMENT_KEY, MJSONWP_ELEMENT_KEY } from '../../protocol/protocol';
+import { ImageElement } from '../image-element';
 
 
 const commands = {}, helpers = {}, extensions = {};
+
+const IMAGE_STRATEGY = "-image";
 
 // Override the following function for your own driver, and the rest is taken
 // care of!
@@ -32,6 +41,15 @@ helpers.findElOrElsWithProcessing = async function (strategy, selector, mult, co
 };
 
 commands.findElement = async function (strategy, selector) {
+  if (strategy === IMAGE_STRATEGY) {
+    // in the singular findElement only, we can find by image, which is
+    // a special case. In any other find command (including
+    // findElement(s)FromElement), this doesn't make sense, so use of the
+    // -image locator strategy won't pass the strategy validation filter as is
+    // desired
+    return await this.findByImage(selector);
+  }
+
   return await this.findElOrElsWithProcessing(strategy, selector, false);
 };
 
@@ -46,6 +64,120 @@ commands.findElementFromElement = async function (strategy, selector, elementId)
 commands.findElementsFromElement = async function (strategy, selector, elementId) {
   return await this.findElOrElsWithProcessing(strategy, selector, true, elementId);
 };
+
+helpers.findByImage = async function (b64Template, {checkingStaleness = false} = {}) {
+  let threshold = DEFAULT_MATCH_THRESHOLD;
+  if (this.settings) {
+    threshold = this.settings.getSettings().imageMatchThreshold;
+  }
+  log.info(`Finding image element with match threshold ${threshold}`);
+  const {width: screenWidth, height: screenHeight} = await this.getWindowSize();
+
+  // someone might have sent in a template that's larger than the screen
+  // dimensions. If so let's check and cut it down to size since the algorithm
+  // will not work unless we do
+  b64Template = await this.ensureTemplateSize(b64Template, screenWidth, screenHeight);
+
+  let rect = null;
+  try {
+    await this.implicitWaitForCondition(async () => {
+      let b64Screenshot = await this.getScreenshotForImageFind(screenWidth, screenHeight);
+      rect = (await this.compareImages(MATCH_TEMPLATE_MODE, b64Screenshot,
+        b64Template, {threshold})).rect;
+      return true;
+    });
+  } catch (err) {
+    if (!err.match(/Condition unmet/)) {
+      throw err;
+    }
+  }
+
+  if (!rect) {
+    throw new errors.NoSuchElementError(); // eslint-disable-line no-unsafe-finally
+  }
+
+  log.info(`Image template matched: ${JSON.stringify(rect)}`);
+  const imgEl = new ImageElement(b64Template, rect);
+
+  // if we're just checking staleness, return straightaway so we don't add
+  // a new element to the cache
+  if (checkingStaleness) {
+    return imgEl;
+  }
+
+  this._imgElCache.add(imgEl);
+  const protoKey = this.isW3CProtocol() ? W3C_ELEMENT_KEY : MJSONWP_ELEMENT_KEY;
+  return imgEl.asElement(protoKey);
+};
+
+helpers.ensureTemplateSize = async function (b64Template, screenWidth, screenHeight) {
+  let {width: tplWidth, height: tplHeight, imgObj} = await getJimpData(b64Template);
+
+  // if the template fits inside the screen dimensions, we're good
+  if (tplWidth <= screenWidth && tplHeight <= screenHeight) {
+    return b64Template;
+  }
+
+  // otherwise, scale it to fit inside the screen dimensions
+  imgObj = imgObj.scaleToFit(screenWidth, screenHeight);
+  return (await imgObj.getBufferPromise(Jimp.MIME_PNG)).toString('base64');
+};
+
+helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
+  let b64Screenshot = await this.getScreenshot();
+  let {width: shotWidth, height: shotHeight, imgObj} = await getJimpData(b64Screenshot);
+
+  if (screenWidth === shotWidth && screenHeight === shotHeight) {
+    // the height and width of the screenshot and the device screen match, which
+    // means we should be safe when doing template matches
+    return b64Screenshot;
+  }
+
+  // otherwise, if they don't match, it could spell problems for the accuracy
+  // of coordinates returned by the image match algorithm, since we match based
+  // on the screenshot coordinates not the device coordinates themselves. There
+  // are two potential types of mismatch: aspect ratio mismatch and scale
+  // mismatch. we need to detect and fix both
+
+  const screenAR = screenWidth / screenHeight;
+  const shotAR = shotWidth / shotHeight;
+
+  if (screenAR !== shotAR) {
+    log.warn(`When trying to find an element, determined that the screen ` +
+             `aspect ratio and screenshot aspect ratio are different. Screen ` +
+             `is ${screenWidth}x${screenHeight} whereas screenshot is ` +
+             `${shotWidth}x${shotHeight}. Resizing screenshot to match ` +
+             `screen aspect ratio so that image element coordinates have a ` +
+             `greater chance of being correct.`);
+    shotWidth = shotWidth / (shotAR / screenAR);
+    imgObj = imgObj.resize(shotWidth, shotHeight);
+  }
+
+  // now we know the aspect ratios match, but there might still be a scale
+  // mismatch, so just resize based on the screen dimensions
+  if (screenWidth !== shotWidth) {
+    imgObj = imgObj.resize(screenWidth, Jimp.AUTO);
+  }
+
+  return (await imgObj.getBufferPromise(Jimp.MIME_PNG)).toString('base64');
+};
+
+function getJimpData (b64Data) {
+  return new Promise((resolve, reject) => {
+    new Jimp(Buffer.from(b64Data, 'base64'), (err, imgObj) => {
+      if (err) {
+        return reject(err);
+      }
+      imgObj.getBufferPromise = Promise.promisify(imgObj.getBuffer.bind(imgObj),
+        {context: imgObj});
+      resolve({
+        width: imgObj.bitmap.width,
+        height: imgObj.bitmap.height,
+        imgObj
+      });
+    });
+  });
+}
 
 Object.assign(extensions, commands, helpers);
 export { commands, helpers};
--- a/lib/basedriver/commands/find.js
+++ b/lib/basedriver/commands/find.js
@@ -84,8 +84,12 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
 helpers.findByImage = async function (b64Template, {
   shouldCheckStaleness = false,
   multiple = false,
-} = {}) {
-  const threshold = this.settings.getSettings().imageMatchThreshold;
+}) {
+  const {
+    imageMatchThreshold: threshold,
+    fixImageTemplateSize
+  } = this.settings.getSettings();
+
   log.info(`Finding image element with match threshold ${threshold}`);
   const {width: screenWidth, height: screenHeight} = await this.getWindowSize();
 
@@ -93,7 +97,7 @@ helpers.findByImage = async function (b64Template, {
   // dimensions. If so let's check and cut it down to size since the algorithm
   // will not work unless we do. But because it requires some potentially
   // expensive commands, only do this if the user has requested it in settings.
-  if (this.settings.getSettings().fixImageTemplateSize) {
+  if (fixImageTemplateSize) {
     b64Template = await this.ensureTemplateSize(b64Template, screenWidth,
       screenHeight);
   }
@@ -113,7 +117,7 @@ helpers.findByImage = async function (b64Template, {
     // _other_ kind of error, it means something blew up totally apart from the
     // implicit wait timeout. We should not mask that error and instead throw
     // it straightaway
-    if (!err.match(/Condition unmet/)) {
+    if (!err.message.match(/Condition unmet/)) {
       throw err;
     }
   }
@@ -122,7 +126,7 @@ helpers.findByImage = async function (b64Template, {
     if (multiple) {
       return [];
     }
-    throw new errors.NoSuchElementError(); // eslint-disable-line no-unsafe-finally
+    throw new errors.NoSuchElementError();
   }
 
   log.info(`Image template matched: ${JSON.stringify(rect)}`);
--- a/lib/basedriver/commands/find.js
+++ b/lib/basedriver/commands/find.js
@@ -63,6 +63,24 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
   return await this.findElOrElsWithProcessing(strategy, selector, true, elementId);
 };
 
+/**
+ * @typedef {Object} FindByImageOptions
+ * @property {boolean} shouldCheckStaleness - whether this call to find an
+ * image is merely to check staleness. If so we can bypass a lot of logic
+ * @property {boolean} multiple - Whether we are finding one element or
+ * multiple
+ */
+
+/**
+ * Find a screen rect represented by an ImageElement corresponding to an image
+ * template sent in by the client
+ *
+ * @param {string} b64Template - base64-encoded image used as a template to be
+ * matched in the screenshot
+ * @param {FindByImageOptions}
+ *
+ * @returns {WebElement} - WebDriver element with a special id prefix
+ */
 helpers.findByImage = async function (b64Template, {
   shouldCheckStaleness = false,
   multiple = false,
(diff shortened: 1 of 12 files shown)


lib/basedriver/commands/find.js
```diff
@@ -1,8 +1,17 @@
 import log from '../logger';
+import Jimp from 'jimp';
+import Promise from 'bluebird';
```
mykola-mokhnach: We usually call it `B`

lib/basedriver/commands/find.js
```diff
@@ -1,8 +1,17 @@
...
+import { errors } from '../../..';
+import { Buffer } from 'buffer';
```
mykola-mokhnach: should this be imported explicitly?

lib/basedriver/commands/find.js
```diff
@@ -32,6 +41,15 @@ helpers.findElOrElsWithProcessing = async function (strategy, selector, mult, co
 };
 
 commands.findElement = async function (strategy, selector) {
+  if (strategy === IMAGE_STRATEGY) {
+    // in the singular findElement only, we can find by image, which is
+    // a special case. In any other find command (including
+    // findElement(s)FromElement), this doesn't make sense, so use of the
+    // -image locator strategy won't pass the strategy validation filter as is
+    // desired
+    return await this.findByImage(selector);
```
mykola-mokhnach: I would say there's still a difference, because in case of multiple elements we need to return an empty array or throw an exception if only one element is expected

lib/basedriver/commands/find.js
```diff
@@ -47,6 +65,120 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
   return await this.findElOrElsWithProcessing(strategy, selector, true, elementId);
 };
 
+helpers.findByImage = async function (b64Template, {checkingStaleness = false} = {}) {
```
mykola-mokhnach: -> shouldCheckStaleness

lib/basedriver/commands/find.js
```diff
@@ -47,6 +65,120 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
...
+  let threshold = DEFAULT_MATCH_THRESHOLD;
+  if (this.settings) {
+    threshold = this.settings.getSettings().imageMatchThreshold;
```
mykola-mokhnach: shouldn't we only assign it if `imageMatchThreshold` is defined?

lib/basedriver/commands/find.js
```diff
@@ -47,6 +65,120 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
...
+  }
+  log.info(`Finding image element with match threshold ${threshold}`);
+  const {width: screenWidth, height: screenHeight} = await this.getWindowSize();
+
+  // someone might have sent in a template that's larger than the screen
+  // dimensions. If so let's check and cut it down to size since the algorithm
+  // will not work unless we do
+  b64Template = await this.ensureTemplateSize(b64Template, screenWidth, screenHeight);
```
mykola-mokhnach: I would rather skip this check and let the client to be responsible for template size. It might be too expensive to call `getWindowSize` each time the lookup is executed.

lib/basedriver/commands/find.js
```diff
@@ -47,6 +65,120 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
...
+  let rect = null;
+  try {
+    await this.implicitWaitForCondition(async () => {
+      let b64Screenshot = await this.getScreenshotForImageFind(screenWidth, screenHeight);
+      rect = (await this.compareImages(MATCH_TEMPLATE_MODE, b64Screenshot,
+        b64Template, {threshold})).rect;
+      return true;
+    });
+  } catch (err) {
+    if (!err.match(/Condition unmet/)) {
```
mykola-mokhnach: It might be handy to have some explanation here

lib/basedriver/commands/find.js
```diff
@@ -47,6 +65,120 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
...
+    return b64Screenshot;
+  }
+
+  // otherwise, if they don't match, it could spell problems for the accuracy
+  // of coordinates returned by the image match algorithm, since we match based
+  // on the screenshot coordinates not the device coordinates themselves. There
+  // are two potential types of mismatch: aspect ratio mismatch and scale
+  // mismatch. we need to detect and fix both
+
+  const screenAR = screenWidth / screenHeight;
```
mykola-mokhnach: screen aspect ration is not going to be changed during the session, so it might be safe to cache it

lib/basedriver/driver.js
```diff
@@ -276,6 +280,11 @@ class BaseDriver extends Protocol {
         // have an unexpected shutdown event, for example, we can cancel it from
         // outside, rejecting the current command immediately
         this.curCommandCancellable = new B((r) => { r(); }).then(() => { // eslint-disable-line promise/prefer-await-to-then
+          // if one of the args is an image element, handle it separately
+          const imgElId = getImgElFromArgs(args);
```
mykola-mokhnach: this looks a bit hacky

lib/basedriver/image-element.js
```diff
@@ -0,0 +1,169 @@
...
+    // where we expect it to be
+    let newImgEl;
+
+    try {
+      newImgEl = await driver.findByImage(this.template, {checkingStaleness: true});
+    } catch (err) {
+      throw new errors.StaleElementReferenceError();
+    }
+
+    if (!this.equals(newImgEl)) {
```
mykola-mokhnach: I'm not quite sure if `click` should mutate the element

lib/basedriver/commands/find.js
```diff
@@ -47,6 +65,120 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
...
+  let b64Screenshot = await this.getScreenshot();
+  let {width: shotWidth, height: shotHeight, imgObj} = await getJimpData(b64Screenshot);
+
+  if (screenWidth === shotWidth && screenHeight === shotHeight) {
+    // the height and width of the screenshot and the device screen match, which
+    // means we should be safe when doing template matches
+    return b64Screenshot;
+  }
+
+  // otherwise, if they don't match, it could spell problems for the accuracy
```
mykola-mokhnach: I would rather make this configurable from settings to make it as performant as possible by default and only do these transformations if explicitly asked by the client

jlipps:
(note that this PR now depends on https://github.com/appium/appium-support/pull/77)

lib/basedriver/commands/find.js
```diff
@@ -111,7 +110,8 @@ helpers.findByImage = async function (b64Template, {checkingStaleness = false} =
 };
 
 helpers.ensureTemplateSize = async function (b64Template, screenWidth, screenHeight) {
-  let {width: tplWidth, height: tplHeight, imgObj} = await getJimpData(b64Template);
+  let imgObj = await imageUtils.getJimpData(b64Template);
```
mykola-mokhnach: is the function called `getJimpImage` or `getJimpData` ?

jlipps: oops, fixed.

lib/basedriver/commands/find.js
jlipps: i think it could be changed during session if the device orientation is changed

lib/basedriver/driver.js
jlipps: hmm, it was the only way I could think of to pass control to ImageElement in the case of an image element existing in the arguments. I agree it's not super pretty. Any other ideas?

lib/basedriver/commands/find.js
imurchie: But in case of multiple elements (or anything other than strict `findElement`) it will throw an error saying there is no such strategy (https://github.com/appium/appium-base-driver/blob/master/lib/basedriver/commands/find.js#L19).

lib/basedriver/commands/find.js
```diff
@@ -64,18 +63,40 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
...
 };
 
-helpers.findByImage = async function (b64Template, {checkingStaleness = false} = {}) {
-  let threshold = DEFAULT_MATCH_THRESHOLD;
-  if (this.settings) {
-    threshold = this.settings.getSettings().imageMatchThreshold;
-  }
+/**
+ * @typedef {Object} FindByImageOptions
+ * @property {boolean} shouldCheckStaleness - whether this call to find an
```
mykola-mokhnach: can we also mention the default values for these properties or if any of them is required to set?

lib/basedriver/commands/find.js
```diff
@@ -64,18 +63,40 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
...
+ * @param {string} b64Template - base64-encoded image used as a template to be
+ * matched in the screenshot
+ * @param {FindByImageOptions}
+ *
+ * @returns {WebElement} - WebDriver element with a special id prefix
+ */
+helpers.findByImage = async function (b64Template, {
+  shouldCheckStaleness = false,
+  multiple = false,
+} = {}) {
```
mykola-mokhnach: I assume `= {}` is redundant

lib/basedriver/commands/find.js
```diff
@@ -64,18 +63,40 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
...
+  const threshold = this.settings.getSettings().imageMatchThreshold;
```
mykola-mokhnach: does getSettings return a promise?

mykola-mokhnach:
```
const {
  imageMatchThreshold: threshold,
  fixImageTemplateSize
} = this.settings.getSettings();
```

lib/basedriver/commands/find.js
```diff
@@ -86,29 +107,49 @@ helpers.findByImage = async function (b64Template, {checkingStaleness = false} =
       return true;
     });
   } catch (err) {
+    // this `implicitWaitForCondition` method will throw a 'Condition unmet'
+    // error if an element is not found eventually. In this case, we will
+    // handle the element not found response below. In the case where get some
+    // _other_ kind of error, it means something blew up totally apart from the
+    // implicit wait timeout. We should not mask that error and instead throw
+    // it straightaway
     if (!err.match(/Condition unmet/)) {
```
mykola-mokhnach: should me match err.message instead?

lib/basedriver/device-settings.js
```diff
@@ -3,7 +3,30 @@ import log from './logger';
...
+  // against the current screenshot before clicking it
+  checkForImageElementStaleness: true,
+
+  // whether before clicking on an image element Appium should re-determine the
+  // position of the element on screen
+  autoUpdateImageElementPosition: false,
+
+  // which method to use for tapping by coordinate for image elements. the
+  // options are 'w3c' or 'mjsonwp'
+  imageElementTapStrategy: 'w3c',
```
mykola-mokhnach: shouldn't this be detected automatically?

lib/protocol/protocol.js
```diff
@@ -439,7 +439,7 @@ function driverShouldDoJwpProxy (driver, req, command) {
 
   // if it looks like we have an image element in the body, never proxy
   const stringBody = JSON.stringify(req.body);
```
mykola-mokhnach: I'm still not quite sure about that. What if any other valid request includes accidentally includes '-image' substring in its body?

lib/basedriver/image-element.js
```diff
@@ -6,76 +6,174 @@ import UUID from 'uuid-js';
...
+      // set up a W3C action to click on the image by position
+      log.info('Will tap using W3C actions');
+      const action = {
+        type: 'pointer',
+        id: 'mouse',
+        parameters: {pointerType: 'touch'},
+        actions: [
+          {type: 'pointerMove', x, y},
+          {type: 'pointerDown'},
+          {type: 'pause', duration: 500},
```
mykola-mokhnach: 500 is the duration of a long tap. single tap duration is about 100-125 ms

lib/basedriver/commands/find.js
jlipps: > does getSettings return a promise?

no

lib/basedriver/commands/find.js
jlipps: yes!

lib/basedriver/device-settings.js
jlipps: how can it be detected automatically? we don't know which functionality the downstream driver supports. i.e., this doesn't (and shouldn't) correspond to the protocol used. it corresponds to the functionality available in the driver.

lib/protocol/protocol.js
jlipps: it would have to include `appium-image-element-`. but you are right it is a risk. for a similar reason the webdriver spec does the whole `element-6066-11e4-a52e-4f735466cecf` thing.

one alternative is actually walking through the body recursively and seeing if we have an object with the exact form of an element. that seemed like it might be overkill. @imurchie what do you think?

mykola-mokhnach: having such suffix might be useful IMHO

lib/basedriver/device-settings.js
mykola-mokhnach: we know in which mode the downstream driver is running (w3c/jsonwp), don't we?

jlipps: I'm not sure that we do. And anyway the mode doesn't matter: all we're doing is calling `driver.performActions` or `driver.performTouch`. There's no protocol involved. I'm just giving access to these methods protocol-related names because that's how they're known. But maybe strings like `w3cAction` and `touchAction` would be better.

But your comment makes me realize that this will not be proxied so in the case of WDA, we'll have to ensure the xcuitest driver has `performActions` on it, and manually proxy the command forward to WDA. That's not ideal.

lib/protocol/protocol.js
```diff
@@ -418,22 +419,14 @@ function driverShouldDoJwpProxy (driver, req, command) {
...
-    if (avoidMethod === req.method && avoidPathRegex.test(normalizedUrl)) {
-      return false;
-    }
+  if (driver.proxyRouteIsAvoided(req.params.sessionId, req.method, req.originalUrl)) {
+    return false;
+  }
+
+  // if it looks like we have an image element in the body, never proxy
+  const stringBody = JSON.stringify(req.body);
+  if (stringBody && stringBody.includes(IMAGE_ELEMENT_PREFIX)) {
```
mykola-mokhnach: wouldn't it be safer to change includes to `/`\b${IMAGE_ELEMENT_PREFIX}`/.test(stringBody)` ?

jlipps:
ok i've fixed all the issues reported so far and written all the unit tests that need writing. i will move on to writing functional tests in the xcui and uiauto2 drivers, so let's wait to merge this until i've at least run them locally. meanwhile i added a few significant commits since the last review so might want to have another look at it.

lib/protocol/protocol.js
```diff
@@ -17,6 +17,16 @@ const LOG_OBJ_LENGTH = 1024; // MAX LENGTH Logged to file / console
 
 const MJSONWP_ELEMENT_KEY = 'ELEMENT';
 const W3C_ELEMENT_KEY = 'element-6066-11e4-a52e-4f735466cecf';
+const IMAGE_ELEMENT_PREFIX = 'appium-image-element-';
+
+const IMG_EL_BODY_RE = new RegExp(
+  `"(${W3C_ELEMENT_KEY}|${MJSONWP_ELEMENT_KEY})": ?` +
```
mykola-mokhnach: perhaps ` ?` -> `\s*`

lib/basedriver/driver.js
```diff
@@ -408,6 +423,31 @@ class BaseDriver extends Protocol {
     return false;
   }
 
+  proxyRouteIsAvoided (sessionId, method, url) {
+    for (let avoidSchema of this.getProxyAvoidList(sessionId)) {
+      if (!_.isArray(avoidSchema) || avoidSchema.length !== 2) {
```
mykola-mokhnach: this will throw inside avoidSchema.length if `avoidSchema` is null

lib/basedriver/driver.js
```diff
@@ -408,6 +423,31 @@ class BaseDriver extends Protocol {
...
+        throw new Error('Proxy avoidance must be a list of pairs');
+      }
+      let [avoidMethod, avoidPathRegex] = avoidSchema;
+      if (!_.includes(['GET', 'POST', 'DELETE'], avoidMethod)) {
+        throw new Error(`Unrecognized proxy avoidance method '${avoidMethod}'`);
+      }
+      if (!(avoidPathRegex instanceof RegExp)) {
```
mykola-mokhnach: perhaps `_.isRegExp` ?

lib/basedriver/driver.js
```diff
@@ -408,6 +423,31 @@ class BaseDriver extends Protocol {
...
+      }
+      let normalizedUrl = url.replace(/^\/wd\/hub/, '');
+      if (avoidMethod === method && avoidPathRegex.test(normalizedUrl)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  async proxyCommand (/*url, method, body*/) {
```
mykola-mokhnach: why do we need to implement it in each driver?

jlipps: each driver does proxying differently (uiauto2 calls `this.uiautomator2.jwproxy` for example, which isn't the case with xcui)

mykola-mokhnach: cannot we just call `click` or `performActions`? Why the base driver should care about proxying?

jlipps: what if a driver doesn't implement `performActions` itself, but merely proxies requests for it to a downstream driver? for example, if uiauto2 server implemented the /actions endpoint, but uiauto2 driver did not. this is a totally valid case. we have many endpoints which are not specified explicitly in a driver and instead we rely on proxying to access the endpoint on a downstream driver.

does that make sense? maybe i'm confused but i think it's a real issue (at least in principle)

mykola-mokhnach: I wouldn't try to outsmart drivers. If some endpoint is not implemented there and it exists in the downstream driver then it would be better and more practical to implement the missing stuff directly there. Adding new requirements to base driver creates unnecessary complexity for the whole infrastructure.

lib/protocol/protocol.js
jlipps: good idea, will update

lib/basedriver/driver.js
jlipps: existing code (moved from protocol) but good point, i'll fix

lib/basedriver/driver.js
jlipps: yep, existing code but will fix

lib/basedriver/driver.js
jlipps: ok. i think you have a good point. the way i am doing it now we will also require drivers to implement the `proxyCommand` method. so it's not less onerous than requiring them to simply implement `performActions` and `performTouch`. i'll do the same thing as i have done elsewhere and simply throw a meaningful error if these methods do not exist on the driver.

if the driver typically proxies the action/touch commands, then it will need to basically implement a one-line function to support image elements, like:

```js
commands.performActions = function (actions) {
  this.proxyCommand('POST', '/actions', actions);
};
```

In most cases this will be bypassed, but in the case of ImageElement actions, it will be called.

lib/basedriver/driver.js
jlipps: actually, this is OK. if avoidSchema is not an array, the condition will short-circuit since the LHS is true, and `avoidSchema.length` will never be evaluated

jlipps:
ok, tests have been written for xcuitest driver and uiauto2 driver (those PRs are up but won't pass until this is merged). from my perspective this is solid (at least for a beta, presumably we'll find more bugs or important behavior suggestions when it's in the wild).

jlipps:
will also work on documentation, btw.

jlipps:
the one build failure is a travis timeout, so I think this is good to go. Any other thoughts @imurchie @mykola-mokhnach? Otherwise I'm ready to merge.

lib/protocol/protocol.js
```diff
@@ -418,22 +428,26 @@ function driverShouldDoJwpProxy (driver, req, command) {
...
-      throw new Error(`Unrecognized proxy avoidance method '${avoidMethod}'`);
-    }
-    if (!(avoidPathRegex instanceof RegExp)) {
-      throw new Error('Proxy avoidance path must be a regular expression');
-    }
-    let normalizedUrl = req.originalUrl.replace(/^\/wd\/hub/, '');
-    if (avoidMethod === req.method && avoidPathRegex.test(normalizedUrl)) {
-      return false;
-    }
+  if (driver.proxyRouteIsAvoided(req.params.sessionId, req.method, req.originalUrl)) {
```
mykola-mokhnach: it might be useful to add a docstring to this new method

jlipps:
@mykola-mokhnach ok docstring added. let me know if you're happy with things as they stand now.

jlipps:
@imurchie how about you?

jlipps:
published in `appium-base-driver@3.4.0`

PR #238, opened 2018-07-21
Try to fix the destination URL if the upstream and downstream drivers operate different protocols

mykola-mokhnach:
I don't like the current implementation very much, but this is the best idea I came to so far.

### DIFF
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -232,15 +232,87 @@ class JWProxy {
     }
   }
 
+  async getDownstreamProtocol () {
+    const [, resBody] = await this.proxy(`${this.scheme}://${this.server}:${this.port}${this.base}/status`, 'GET');
+    return this.getProtocolFromResBody(resBody);
+  }
+
+  async proxyCommand (url, method, body = null) {
+    // Handle "crossing" endpoints for the case
+    // when upstream and downstream drivers operate different protocols
+
+    // Same url, but different arguments
+
+    if (method.toLowerCase() === 'post' && url.match(/\/timeouts$/))  {
+      return await this.proxySetTimeouts(url, method, body);
+    }
+
+    // Same arguments, but different URLs
+
+    const proxyToFixedEndpoint = async (urlPatterns, expectedMethod, jsonwpUrlReplacer, w3cUrlReplacer) => {
+      if (method.toLowerCase() !== expectedMethod || !_.some(urlPatterns, (x) => url.match(x))) {
+        return;
+      }
+
+      let downstreamProtocol;
+      try {
+        log.info(`The '${url}' API might not be the same for the downstream protocol. ` +
+          'Trying to proxy as is');
+        return await this.proxy(url, method, body);
+      } catch (e) {
+        try {
+          downstreamProtocol = await this.getDownstreamProtocol();
+        } catch (err) {
+          log.warn(`The downstream protocol cannot be detected. ` +
+            `Original error: ${err.message}`);
+        }
+        if (!downstreamProtocol) {
+          throw e;
+        }
+      }
+      if (downstreamProtocol === MJSONWP) {
+        const rewrittenUrl = jsonwpUrlReplacer(url);
+        log.info(`Rewriting the URL to ${rewrittenUrl}`);
+        return await this.proxy(rewrittenUrl, method, body);
+      }
+      if (downstreamProtocol === W3C) {
+        const rewrittenUrl = w3cUrlReplacer(url);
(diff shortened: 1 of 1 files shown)


lib/jsonwp-proxy/proxy.js
```diff
@@ -232,15 +233,108 @@ class JWProxy {
...
+      if (!commandNames.includes(commandName)) {
+        continue;
+      }
+
+      try {
+        log.info(`The '${url}' API might not be the same for the downstream protocol. ` +
+          'Trying to proxy as is');
+        return await this.proxy(url, method, body);
+      } catch (e) {
+        let downstreamProtocol;
```
imurchie: Are there cases where the downstream protocol changes for a given proxy?

mykola-mokhnach: Practically not, but theoretically it is possible. Which approach do you like more?

imurchie: If it doesn't happen, either the protocol should be saved the first time this code is run, or the proxy should be retrieved through a factory function (so it can be async) which does the single call to figure out the protocol (or, if we know the protocol beforehand, takes the protocol and doesn't do the check), and saves it, so every call can be rewritten correctly.

mykola-mokhnach: ok, I'm caching it now

lib/jsonwp-proxy/proxy.js
```diff
@@ -232,15 +236,111 @@ class JWProxy {
...
+    return this._downstreamProtocol;
+  }
+
+  requestToCommandName (url, method) {
+    const extractCommandName = (pattern) => {
+      const pathMatch = pattern.exec(url);
+      return pathMatch ? routeToCommandName(pathMatch[1], method) : null;
+    };
+    let commandName = routeToCommandName(url, method);
+    if (!commandName && _.includes(url, '/wd/hub/session/')) {
```
jlipps: what if we move to urls which don't include `/wd/hub` as a prefix? this isn't part of the spec so we could conceivably move to no prefix, or configurable prefix

lib/jsonwp-proxy/proxy.js
```diff
@@ -232,15 +236,111 @@ class JWProxy {
...
+
+    // Same url, but different arguments
+
+    if (commandName === 'timeouts')  {
+      return await this.proxySetTimeouts(url, method, body);
+    }
+
+    // Same arguments, but different URLs
+
+    const conflicts = [
```
jlipps: i don't care much one way or another, but this could be a file-level constant (and the embedded functions would then just take a url param)

lib/jsonwp-proxy/proxy.js
```diff
@@ -232,15 +236,111 @@ class JWProxy {
...
+          '/element/$1/screenshot'),
+      },
+    ];
+
+    for (const {commandNames, jsonwpConverter, w3cConverter} of conflicts) {
+      if (!commandNames.includes(commandName)) {
+        continue;
+      }
+
+      try {
```
jlipps: why do we try the proxy first not knowing whether it will work? could this conceivably have side effects? and wouldn't we only want to run the catch block in the case of a protocol mismatch error? how are we filtering out all other kinds of errors?

i think we could safely check the downstream protocol once for each session, and then just save it. then we simply cross protocols if we know we need to.

lib/jsonwp-proxy/proxy.js
```diff
@@ -232,15 +236,111 @@ class JWProxy {
...
+    log.debug(`Matched '${url}' to command name '${commandName}'`);
+
+    // Handle "crossing" endpoints for the case
+    // when upstream and downstream drivers operate different protocols
+
+    // Same url, but different arguments
+
+    if (commandName === 'timeouts')  {
+      return await this.proxySetTimeouts(url, method, body);
+    }
```
jlipps: isn't send keys also one of these with the same url but different arguments?

lib/jsonwp-proxy/proxy.js
mykola-mokhnach: `routeToCommandName` call itself should cover such case. These two conditions were added just for the sake of safety.

lib/jsonwp-proxy/proxy.js
mykola-mokhnach: yes, it is easy to do that.

lib/jsonwp-proxy/proxy.js
mykola-mokhnach: for now we don't see any problems with sendKeys so far. I'll consider adding it as soon as there are any issues reported

PR #239, opened 2018-07-24
Fix handling of proxy url

imurchie:
Use [lazy quantifier](https://javascript.info/regexp-greedy-and-lazy#lazy-mode) in proxy url processing, so we don't consume the whole url.

This does what #236 failed to do.

### DIFF
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -72,7 +72,7 @@ class JWProxy {
       throw new Error(`Did not know what to do with url '${url}'`);
     }
 
-    const stripPrefixRe = new RegExp('^.+(/(session|status).*)$');
+    const stripPrefixRe = new RegExp('^.+?(/(session|status).*)$');
     if (stripPrefixRe.test(remainingUrl)) {
       remainingUrl = stripPrefixRe.exec(remainingUrl)[1];
     }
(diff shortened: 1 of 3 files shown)


imurchie:
Tested against full `appium-xcuitest-driver` suite with no failures of the sort reported from #236.

lib/jsonwp-proxy/proxy.js
```diff
@@ -72,7 +72,7 @@ class JWProxy {
       throw new Error(`Did not know what to do with url '${url}'`);
     }
 
-    const stripPrefixRe = new RegExp('^.+(/(session|status).*)$');
+    const stripPrefixRe = new RegExp('^.+?(/(session|status).*)$');
```
mykola-mokhnach: I don't quite understand what exactly we are trying to match here. Isn't `.+?` the same as `.*` ?

imurchie: No. 

* `.*` matches zero or more of any character, greedily (that is, getting as many characters as possible until it reaches some other condition that is fulfilled obligately)
* `.+?` matches one or more of any character, lazily (that is, getting as _few_ characters as possible but still fulfilling obligate conditions)


But you've pointed out a valid thing. This should be `^.*?` not `^.+?` to handle the case where there is nothing in front of the `/session`, though that case is also not handled in the current code, either.

So, taking something simple like the sentence `"word is a word"`,
* `.*?(word)` will retrieve the first instance of `word`
* `.*(word)` will retrieve the second instance of `word`

mykola-mokhnach: thanks for the detailed explanation

imurchie: Of course!

PR #241, opened 2018-07-25
Get message for MJSONWP error

imurchie:
Recent changes have broken the handling of proxy errors in certain cases. In particular, Chromedriver returns a value that has a message inside it, so we need to unwrap it.

See https://github.com/appium/appium-chromedriver/issues/90

### DIFF
--- a/lib/protocol/errors.js
+++ b/lib/protocol/errors.js
@@ -771,6 +771,9 @@ function isErrorType (err, type) {
  * @return {ProtocolError} The error that is associated with provided JSONWP status code
  */
 function errorFromMJSONWPStatusCode (code, message) {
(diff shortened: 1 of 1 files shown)


lib/protocol/errors.js
```diff
@@ -771,6 +771,9 @@ function isErrorType (err, type) {
  * @return {ProtocolError} The error that is associated with provided JSONWP status code
  */
 function errorFromMJSONWPStatusCode (code, message) {
+  if (message && message.message) {
```
mykola-mokhnach: it might be handy to have some explanation comment here

mykola-mokhnach:
^ :-D

PR #242, opened 2018-07-25
Add method to check for unknown error

imurchie:


lib/protocol/errors.js
```diff
@@ -742,7 +742,12 @@ for (let ErrorClass of _.values(errors)) {
   }
 }
 
-
+function isUnknownError (err) {
+  return !err.constructor.name ||
+         !_.keys(errors).find(function (error) {
```
mykola-mokhnach: you could probably use `_.some`

lib/protocol/protocol.js
```diff
@@ -352,7 +354,7 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
       if (isErrorType(err, errors.ProxyRequestError)) {
         protocolLogger.error(`Encountered internal error running command:  ${JSON.stringify(err)} ${err.stack}`);
         actualErr = err.getActualError();
-      } else if (!isW3C && (!(isErrorType(err, ProtocolError) || isErrorType(err, errors.BadParametersError)))) {
+      } else if (!isW3C && isUnknownError(err)) {
```
mykola-mokhnach: I still don't quite understand why `!isW3C` is necessary here

mykola-mokhnach: perhaps @dpgraham can explain

lib/protocol/protocol.js
```diff
@@ -352,7 +354,7 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
...
         protocolLogger.error(`Encountered internal error running command: ${err.stacktrace || err.stack}`);
         actualErr = new errors.UnknownError(err);
```
mykola-mokhnach: can we now revert the change in `lib/jsonwp-proxy/proxy.js`? Perhaps the whole try/catch is not needed there now

lib/protocol/protocol.js
```diff
@@ -352,7 +354,7 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
       if (isErrorType(err, errors.ProxyRequestError)) {
```
mykola-mokhnach: maybe it might be handy to just use duck typing here and check if the err object has the corresponding method/property

lib/protocol/protocol.js
imurchie: I have no idea. It was there, so I'm not ready to remove.

lib/protocol/protocol.js
imurchie: It can't be reverted, or _every_ proxied error will turn into an unknown error (see https://github.com/appium/appium-base-driver/blob/master/lib/protocol/protocol.js#L452-L458). Removing the whole catch in https://github.com/appium/appium-base-driver/blob/master/lib/jsonwp-proxy/proxy.js#L399-L405 might be possible.

mykola-mokhnach: yes, lets remove the whole

lib/protocol/protocol.js
dpgraham: The `getResponseForW3CError` that's called afterwards covers the conversions so it was unnecessary.

mykola-mokhnach: lets add an explanation comment here then

lib/protocol/errors.js
```diff
@@ -742,7 +742,12 @@ for (let ErrorClass of _.values(errors)) {
   }
 }
 
-
+function isUnknownError (err) {
+  return !err.constructor.name ||
```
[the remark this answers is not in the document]
imurchie: It should _always_ be there (so should `name`, really). Even for bare objects:
```js
$ node
> a = {};
{}
> a.constructor
[Function: Object]
> a.constructor.name
'Object'
>
```

PR #244, opened 2018-08-01
Add an optional parameter to /window call

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/10319

### DIFF
--- a/lib/protocol/routes.js
+++ b/lib/protocol/routes.js
@@ -105,7 +105,8 @@ const METHOD_MAP = {
     GET: {command: 'getWindowHandle'},
     POST: {command: 'setWindow', payloadParams: {
       optional: ['name', 'handle'],
-      validate: (jsonObj) => (!jsonObj.name && !jsonObj.handle) && 'we require one of "name" or "handle" to be set',
+      validate: (jsonObj) => (!util.hasValue(jsonObj.name) && !util.hasValue(jsonObj.handle))
+        && 'we require one of "name" or "handle" to be set',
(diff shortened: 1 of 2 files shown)


lib/protocol/routes.js
```diff
@@ -103,7 +103,10 @@ const METHOD_MAP = {
   },
   '/wd/hub/session/:sessionId/window': {
     GET: {command: 'getWindowHandle'},
-    POST: {command: 'setWindow', payloadParams: {required: ['name']}},
+    POST: {command: 'setWindow', payloadParams: {
```
jlipps: really what we mean is that either name or handle may be present, but at least one _must_ be. is there a way to express that here?

imurchie: Where is the `name` coming from? The spec seems to just want a mandatory `handle`.

mykola-mokhnach: I'm not sure where the `name` comes from. It just was there. I could change the current implementation and add a validation helper, which is gonna check if either parameter is set

jlipps: `name` is [JSONWP](https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidwindow), `handle` is W3C

jlipps: this is basically the same situation as sendKeys

mykola-mokhnach: ok, I've added makeArgs call similar to sendKeys

PR #246, opened 2018-08-09
Support `pageLoadStrategy` capability

JonathanWylie:
This capability is now supported by chromedriver.

### DIFF
--- a/lib/basedriver/desired-caps.js
+++ b/lib/basedriver/desired-caps.js
@@ -52,6 +52,9 @@ let desiredCapabilityConstraints = {
   printPageSourceOnFindFailure: {
     isBoolean: true
   },
+  pageLoadStrategy: {
+    isString: true
+  }
 };
 
 validator.validators.isString = function (value) {


lib/basedriver/desired-caps.js
```diff
@@ -52,6 +52,9 @@ let desiredCapabilityConstraints = {
   printPageSourceOnFindFailure: {
     isBoolean: true
   },
+  pageLoadStrategy: {
```
mykola-mokhnach: should it be defined in base driver or android driver instead?

imurchie: Definitely Android driver, since this is not available on iOS. Since the capability will need to be added to the handling of Chromedriver anyway, it an just be part of that work.

https://github.com/appium/appium-android-driver/blob/master/lib/desired-caps.js

Thanks for the update, though!

JonathanWylie: Your right it's not strictly required to make it work.
BUT if it is not in the basedriver then it logs a complaint.
```
[BaseDriver] The following capabilities were provided, but are not recognized by appium: pageLoadStrategy.
```

JonathanWylie: Maybe basedriver should not complain about unrecognized capabilities, if it is not supposed to know what all the capabilities are.

JonathanWylie: Any response?

mykola-mokhnach: I assume for now the best place where this cap could be defined is https://github.com/appium/appium-android-driver/blob/8ec5e67008a20e27dee3c48958c77cfc3d2e6d6d/lib/desired-caps.js#L174

imurchie: If it is in the `appium-android-driver` there should be no complaints for it _or_ the `appium-uiautomator2-driver`.

JonathanWylie: As requested, I have made an alternative pull request to appium-base-driver: https://github.com/appium/appium-android-driver/pull/409

PR #251, opened 2018-08-21
Add JSONWP<>W3C mapping for window/handles endpoint

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/11219


dpgraham:
Nice, thanks for staying on top of the W3C stuff.

PR #254, opened 2018-09-06
Fix setting alert text for W3C protocol

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/11323


dpgraham:
Published to 3.6.1

PR #256, opened 2018-09-07
isError lodash verification seems to be not very reliable for us

mykola-mokhnach:
Addresses https://gist.github.com/jlegorburuJC/ca0ee8b2f724c224f4da1bc590bae5f1#file-appium-log-L183


mykola-mokhnach:
The fix has been verified by the issue reporter

KazuCocoa:
👍

PR #257, opened 2018-09-07
Fix protocol detection inside the routing function

mykola-mokhnach:
I hope this will address https://github.com/appium/appium/issues/11300

### DIFF
--- a/lib/protocol/protocol.js
+++ b/lib/protocol/protocol.js
@@ -7,10 +7,11 @@ import { METHOD_MAP, NO_SESSION_ID_COMMANDS } from './routes';
 import { renameKey } from '../basedriver/helpers';
 import B from 'bluebird';
 import BaseDriver from '../basedriver/driver';
-
+import LRU from 'lru-cache';
 
 const mjsonwpLog = logger.getLogger('MJSONWP');
 const w3cLog = logger.getLogger('W3C');
+const genericProtocolLog = logger.getLogger('GENERIC');
 
 const JSONWP_SUCCESS_STATUS_CODE = 0;
 // TODO: Make this value configurable as a server side capability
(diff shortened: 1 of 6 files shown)


lib/protocol/protocol.js
```diff
@@ -31,8 +31,65 @@ const IMG_EL_URL_RE = new RegExp(
...
+    default:
+      return genericProtocolLog;
+  }
+}
+
+// This cache is useful when a session gets terminated
+// and removed from the sessions list in the umbrella driver,
+// but the client still tries to send a command to the same session id.
+// So we know how to properly wrap the error message for it
+const PROTOCOLS_MAPPING_CACHE = {};
```
jlipps: we already use an lru cache module for other purposes, perhaps it could simplify this?

lib/protocol/protocol.js
```diff
@@ -31,8 +31,65 @@ const IMG_EL_URL_RE = new RegExp(
...
+      return value;
+    }
+    if (PROTOCOLS_MAPPING_CACHE[sessionId]) {
+      return PROTOCOLS_MAPPING_CACHE[sessionId];
+    }
+    if (!value) {
+      return null;
+    }
+
+    PROTOCOLS_MAPPING_CACHE[sessionId] = [new Date().getTime(), value];
```
jlipps: i usually just use `Date.now()` cause it's short

lib/protocol/protocol.js
```diff
@@ -8,9 +8,9 @@ import { renameKey } from '../basedriver/helpers';
 import B from 'bluebird';
 import BaseDriver from '../basedriver/driver';
 
-
 const mjsonwpLog = logger.getLogger('MJSONWP');
 const w3cLog = logger.getLogger('W3C');
+const genericProtocolLog = logger.getLogger('GENERIC');
```
jlipps: should this be `GENERIC` or `UNKNOWN_PROTOCOL`?

lib/protocol/protocol.js
mykola-mokhnach: I've almost forgotten about it (considering it was my own suggestion :) ) 
Updated now.

lib/protocol/protocol.js
mykola-mokhnach: this is now obsolete

lib/protocol/protocol.js
mykola-mokhnach: I prefer GENERIC, because we include data for both protocols into the response if the protocol is neither W3C nor JSONWP

lib/protocol/protocol.js
jlipps: ha!

dpgraham:
Looks good now. Will merge _pending_ CI passing.

PR #259, opened 2018-09-16
Append the current session identifier to the protocol name prefix

mykola-mokhnach:
This should simplify debugging of parallel drivers

### DIFF
--- a/lib/protocol/protocol.js
+++ b/lib/protocol/protocol.js
@@ -41,7 +41,7 @@ function buildLogger (protocol, sessionId = null) {
     return SESSIONS_CACHE.has(sessionId)
       ? SESSIONS_CACHE.get(sessionId).logger
       : logger.getLogger(`${protocol || GENERIC_PROTOCOL} ` +
-          `(${sessionId.substring(0, sessionId.length >= 8 ? 8 : sessionId.length)})`);
+          `(${sessionId.substring(0, Math.min(sessionId.length, 8))})`);
   }
 
   switch (protocol) {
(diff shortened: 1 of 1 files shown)


lib/protocol/protocol.js
```diff
@@ -35,31 +36,61 @@ const IMG_EL_URL_RE = new RegExp(
...
+  getLogger (sessionId, protocol) {
+    if (sessionId) {
+      if (this._cache.has(sessionId)) {
+        const value = this._cache.get(sessionId);
+        if (value.logger) {
+          return value.logger;
+        }
+        protocol = protocol || value.protocol;
+      }
+      return logger.getLogger(`${protocol || GENERIC_PROTOCOL} ` +
```
imurchie: Should this get added to `value`? If, for whatever reason, there is no logger for a session id, each log line will create a new logger, which is then thrown out.

mykola-mokhnach: This was done intentionally to include the "wrong" session identifier into the actual log lines without caching the whole logger. 

Although, if we don't want to create a new instance each time this method is called, then we could either fall back to the default logger with protocol name only or include some generic prefix there, like `[W3C (unknown session)]`.

mykola-mokhnach: We don't want to put each random session id into the cache in order to save memory. Instead we only put these ids that are returned by `createSession` call

imurchie: Ok. This should be documented, or else it will be "fixed" at some later date.

mykola-mokhnach: I'll add a comment

PR #268, opened 2018-10-02
add custom locator strategy

jlipps:
The idea behind this PR is to open up the possibility of "element finding plugins", which are third-party modules that users can use for element finding functionality that probably shouldn't be maintained by the core project but which nonetheless might be useful. (For example, I'm working on an experimental locator strategy involving a machine-learning model, which probably shouldn't become part of Appium core, at least not for some time).

The idea for how this should be used is as follows:

1. A third party creates a node module that exports a `find` method matching the signature described in this PR. Essentially, that third party module is passed a driver object so it can mix and match Appium capabilities with its own logic.
2. The user takes advantage of this plugin by setting the `customFindModule` capability to the name of the node module.

Right now, we assume that the node module is installed in Appium's tree (or can otherwise be `require`d). In the future, with the Appium 2.0 architecture, we could allow plugins like these to be managed via the Appium CLI.

We might have other kinds of plugins in the future, beyond just element finding ones, but this is the first need I've come across.

Anyway, have a look and once everyone is a fan of the idea I'll work on tests etc.

### DIFF
--- a/lib/basedriver/commands/find.js
+++ b/lib/basedriver/commands/find.js
@@ -1,14 +1,16 @@
 import log from '../logger';
+import { logger, imageUtil } from 'appium-support';
+import _ from 'lodash';
 import { errors } from '../../..';
 import { MATCH_TEMPLATE_MODE } from './images';
 import { W3C_ELEMENT_KEY, MJSONWP_ELEMENT_KEY } from '../../protocol/protocol';
 import { ImageElement } from '../image-element';
-import { imageUtil } from 'appium-support';
 
 
 const commands = {}, helpers = {}, extensions = {};
 
 const IMAGE_STRATEGY = "-image";
+const CUSTOM_STRATEGY = "-custom";
 
 // Override the following function for your own driver, and the rest is taken
 // care of!
@@ -41,6 +43,8 @@ helpers.findElOrElsWithProcessing = async function (strategy, selector, mult, co
 commands.findElement = async function (strategy, selector) {
   if (strategy === IMAGE_STRATEGY) {
     return await this.findByImage(selector, {multiple: false});
+  } else if (strategy === CUSTOM_STRATEGY) {
+    return await this.findByCustom(selector, false);
   }
 
   return await this.findElOrElsWithProcessing(strategy, selector, false);
@@ -49,6 +53,8 @@ commands.findElement = async function (strategy, selector) {
 commands.findElements = async function (strategy, selector) {
   if (strategy === IMAGE_STRATEGY) {
     return await this.findByImage(selector, {multiple: true});
+  } else if (strategy === CUSTOM_STRATEGY) {
+    return await this.findByCustom(selector, true);
   }
 
   return await this.findElOrElsWithProcessing(strategy, selector, true);
@@ -60,6 +66,74 @@ commands.findElementFromElement = async function (strategy, selector, elementId)
 
 commands.findElementsFromElement = async function (strategy, selector, elementId) {
   return await this.findElOrElsWithProcessing(strategy, selector, true, elementId);
+};
+
+/**
+ * Find an element using a custom plugin specified by the customFindModule cap.
+ *
+ * @param {string} selector - the selector which the plugin will use to find
+ * elements
+ * @param {boolean} multiple - whether we want one element or multiple
+ *
+ * @returns {WebElement} - WebDriver element or list of elements
+ */
+commands.findByCustom = async function (selector, multiple) {
+  if (!this.opts.customFindModule) {
+    throw new Error("Finding an element using a pluagin is currently an " +
+                    "incubating feature. To use it you must manually install a " +
+                    "plugin module inside your Appium node_modules tree, " +
+                    "and set the capability 'customFindModule' to the name of " +
+                    "that module so that it can be required.");
+  }
+
+  let finder;
+  try {
+    finder = require(this.opts.customFindModule);
+  } catch (err) {
+    throw new Error(`Could not load your custom find module. Original error: ${err}`);
+  }
+
+  if (!_.isFunction(finder.find)) {
+    throw new Error("Your custom find module did not appear to be constructed " +
+                    "correctly. It needs to export an object with a `find` method.");
+  }
+
+  const customFinderLog = logger.getLogger(this.opts.customFindModule);
+
+  let els;
+  const condition = async () => {
+    // get a list of matched elements from the custom finder, which can
+    // potentially use the entire suite of methods the current driver provides.
+    // the finder should always return a list of elements, but may use the
+    // knowledge of whether we are looking for one or many to perform internal
+    // optimizations
+    els = await finder.find(this, customFinderLog, selector, multiple);
+
+    // if we're looking for multiple elements, or if we're looking for only
+    // one and found it, we're done
+    if (els.length || multiple) {
+      return true;
+    }
+
+    // otherwise we should retry, so return false to trigger the retry loop
+    return false;
+  };
+
+  try {
+    // make sure we respect implicit wait
+    await this.implicitWaitForCondition(condition);
+  } catch (err) {
+    if (err.message.match(/Condition unmet/)) {
+      throw new errors.NoSuchElementError();
+    }
+    throw err;
+  }
+
+  if (multiple) {
+    return els;
+  }
+
+  return els[0];
 };
 
 /**
--- a/lib/basedriver/commands/find.js
+++ b/lib/basedriver/commands/find.js
@@ -79,7 +79,7 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
  */
 commands.findByCustom = async function (selector, multiple) {
   if (!this.opts.customFindModule) {
-    throw new Error("Finding an element using a pluagin is currently an " +
+    throw new Error("Finding an element using a plugin is currently an " +
                     "incubating feature. To use it you must manually install a " +
                     "plugin module inside your Appium node_modules tree, " +
                     "and set the capability 'customFindModule' to the name of " +
@@ -90,28 +90,29 @@ commands.findByCustom = async function (selector, multiple) {
   try {
     finder = require(this.opts.customFindModule);
   } catch (err) {
-    throw new Error(`Could not load your custom find module. Original error: ${err}`);
+    throw new Error(`Could not load your custom find module ` +
+                    `'${this.opts.customFindModule}'. Original error: ${err}`);
   }
 
-  if (!_.isFunction(finder.find)) {
+  if (!finder || !_.isFunction(finder.find)) {
     throw new Error("Your custom find module did not appear to be constructed " +
                     "correctly. It needs to export an object with a `find` method.");
   }
 
   const customFinderLog = logger.getLogger(this.opts.customFindModule);
 
-  let els;
+  let elements;
   const condition = async () => {
     // get a list of matched elements from the custom finder, which can
     // potentially use the entire suite of methods the current driver provides.
     // the finder should always return a list of elements, but may use the
     // knowledge of whether we are looking for one or many to perform internal
     // optimizations
-    els = await finder.find(this, customFinderLog, selector, multiple);
+    elements = await finder.find(this, customFinderLog, selector, multiple);
 
     // if we're looking for multiple elements, or if we're looking for only
     // one and found it, we're done
-    if (els.length || multiple) {
+    if (!_.isEmpty(elements) || multiple) {
       return true;
     }
 
@@ -129,11 +130,7 @@ commands.findByCustom = async function (selector, multiple) {
     throw err;
   }
 
-  if (multiple) {
-    return els;
-  }
-
-  return els[0];
+  return multiple ? elements : elements[0];
 };
 
 /**
--- a/lib/basedriver/commands/find.js
+++ b/lib/basedriver/commands/find.js
@@ -371,5 +371,5 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
 
 
 Object.assign(extensions, commands, helpers);
-export { commands, helpers, IMAGE_STRATEGY };
+export { commands, helpers, IMAGE_STRATEGY, CUSTOM_STRATEGY };
 export default extensions;
--- a/lib/basedriver/commands/find.js
+++ b/lib/basedriver/commands/find.js
@@ -69,7 +69,7 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
 };
 
 /**
- * Find an element using a custom plugin specified by the customFindModule cap.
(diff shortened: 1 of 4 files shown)


lib/basedriver/commands/find.js
```diff
@@ -62,6 +68,74 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
...
+ * Find an element using a custom plugin specified by the customFindModule cap.
+ *
+ * @param {string} selector - the selector which the plugin will use to find
+ * elements
+ * @param {boolean} multiple - whether we want one element or multiple
+ *
+ * @returns {WebElement} - WebDriver element or list of elements
+ */
+commands.findByCustom = async function (selector, multiple) {
+  if (!this.opts.customFindModule) {
```
mykola-mokhnach: do we allow only one such module to be set at the same time or many of them?

lib/basedriver/commands/find.js
```diff
@@ -62,6 +68,74 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
...
+    throw new Error("Finding an element using a pluagin is currently an " +
```
mykola-mokhnach: typo: plugin

lib/basedriver/commands/find.js
```diff
@@ -62,6 +68,74 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
...
+                    "incubating feature. To use it you must manually install a " +
+                    "plugin module inside your Appium node_modules tree, " +
```
mykola-mokhnach: should we explicitly say where to install it? npm might change this in the future though

lib/basedriver/commands/find.js
```diff
@@ -62,6 +68,74 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
...
+  }
+
+  let finder;
+  try {
+    finder = require(this.opts.customFindModule);
+  } catch (err) {
+    throw new Error(`Could not load your custom find module. Original error: ${err}`);
+  }
+
+  if (!_.isFunction(finder.find)) {
```
mykola-mokhnach: `finder && !_.isFunction ..`. would be safer

KazuCocoa:
> This has the disadvantages of requiring a cap in addition to the element call, and allowing only one find plugin to be in use in a given session.

Can we relax the condition by adding the caps as an array like `['my-find-module1', 'my-find-module2']`? But in the case, `mySelector` in _Call findElement('-custom', 'mySelector')_ should be unique in all included modules to avoid conflict each selector name..
This way is implicit. If a user specifies many custom modules, understanding which selector belongs to which module in all modules is probably difficult and complicated.

Personally, I like below way to make managing them easier. (More explicit way)
> driver.findElement('-custom', 'my-find-module:mySelector')

---

If we'll allow **only one custom module** against a session, the original way, `driver.findElement('-custom', 'my-find-module:mySelector')`,  is better though.

dpgraham:
I like the idea but I'm concerned about the practice of inserting a node module into Appium's existing node_modules because that may require tampering with the versions that are in the node_modules that were set by the shrinkwrap.

May I suggest that the plugins be standalone node packages that sit somewhere on the filesystem, either in some permanent directory that stores plugins globally or in a `plugins` directory that's in the same directory as Appium (ignored by npm and git). The plugins can be required like  `require('/path/to/finder-module.js')` and dependencies are injected into the package.

Something like:

```javascript
// magic-finder-plugin.js
function pluginInitializer(appiumDepOne, appiumDepTwo, appiumDepThree, ...) {
   ...
}

export {pluginInitializer};

// finder-plugins.json
[
  {name: 'magic-finder-plugin', path: '/path/to/magic-finder-plugin.js'}
]

// appium-base-driver.js
function initializeFinderPlugins() {
  const plugins = require('./plugins.json');
  for (let plugin of plugins) {
    const {pluginInitializer} = require(pathToFinderPlugin);
    finderPlugin(appiumDepOne, appiumDepTwo, appiumDepThree, ...);
  }
}
```

dpgraham:
What if we made the plugins a server-flag and when you start the server it installs the plugins (if they're not already there)

lib/basedriver/commands/find.js
jlipps: the way the code is written now, just one module. but see the discussion we're having about whether we want something else.

jlipps:
Thanks for the thoughts so far @mykola-mokhnach @KazuCocoa @dpgraham. There are now two design questions on the table, each with a few proposed answers:

1. How do we provide access to custom find functionality?
    a. Through a `customFindModule` cap in conjunction with `-custom` loc strat
    b. Through a `-custom` loc strat in conjunction with a complex selector of the form `<module>:selector`
2. How do we support the finding and loading of plugins?
    a. We do not support it. We simply `require` the named module and force the user to make it require-able by Appium (whether by putting it into Appium's tree, or making it globally available)
    b. Have the user give us two pieces of info about the module: its name and its location. This would allow us to `require` it without it being in Appium's tree. The user is still in charge of installing it
    c. Have the user give us the name and version spec of the module, and _we_ put it somewhere and manage it for them.

As for design question (1), it seems like the consensus so far is that (1a) would be potentially too restrictive, and a more flexible approach would be better so that we could choose from a number of plugins. Are there any other proposals than (1b)?

As for design question (2), the proposals raise some interesting considerations. I went initially with (2a) because anything else encroaches on the territory taken up by [Appium 2.0](https://gist.github.com/jlipps/651b62316603400cabc95ff0f9faf70f). I wanted to be careful not to design something that we would just throw away when we finally decide on an implementation for Appium 2.0.

Alternatively, we could use this as an opportunity to actually move forward with the Appium 2.0 design in an area which isn't super crucial, namely this new plugin feature. We would just need to put the time into making good decisions. At a first pass, something like the `appium-drivers.json` in the Appium 2 proposal makes sense, only expanding it to account for both "drivers" and "plugins" (maybe in the first instance we just worry about plugins, and then later port driver management to the same system).

As for your specific suggestion, @dpgraham:

> What if we made the plugins a server-flag and when you start the server it installs the plugins (if they're not already there)

I like this suggestion a lot, however if we are moving to a CLI installer model for Appium 2.0 (`appium driver install xcuitest`), then I would want to do the same thing for plugins (`appium plugin install custom-element-finder@1.2.0`). (Of course, there's no reason we couldn't _in addition_ have a server flag that auto-installs modules on server start: `appium --plugins=custom-element-finder@1.2.0`)

I'm open here; just wanted to start out with something simple that advanced users could play around with while we determined whether this whole plugin thing was truly useful. Thoughts? Would love to hear @imurchie's views as well.

lib/basedriver/commands/find.js
jlipps: hmm, i guess it's possible that a module is not truthy?

lib/basedriver/commands/find.js
```diff
@@ -62,6 +68,74 @@ commands.findElementsFromElement = async function (strategy, selector, elementId
...
+    throw new Error("Finding an element using a pluagin is currently an " +
+                    "incubating feature. To use it you must manually install a " +
+                    "plugin module inside your Appium node_modules tree, " +
+                    "and set the capability 'customFindModule' to the name of " +
+                    "that module so that it can be required.");
+  }
+
+  let finder;
+  try {
+    finder = require(this.opts.customFindModule);
```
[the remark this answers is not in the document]
jlipps: i'm not sure what else we can check. it's the users responsible to vet 3rd-party plugins, which could indeed be malicious.

probably at some point we'll have a registry of 'approved' plugins that we have vetted.

dpgraham:
Lgrm

jlipps:
@dpgraham @KazuCocoa @mykola-mokhnach just pushed a change that will allow users to register multiple element finder plugins as an object, allowing for a 'shortcut' name. for example, i could register one like this:

```json
{"customFindModules": {"1": "/really/long/path/to/find/plugin/module/"}}
```

and then in my test code:

```js
driver.findElement('-custom', '1:selector');
```

so I can just use `1` as my local alias for that particular plugin. (we also handle the case where there is only one plugin registered, in which case a prefix is not required).

take a look at the code, and i'll start working on unit tests.

jlipps:
tests pushed

test/basedriver/commands/find-specs.js
```diff
@@ -191,3 +197,74 @@ describe('finding elements by image', function () {
...
+  });
+  it('should give a hint to the plugin about whether multiple are requested', async function () {
+    const d = new BaseDriver();
+    d.opts.customFindModules = {f: CUSTOM_FIND_MODULE};
+    await d.findElement(CUSTOM_STRATEGY, "f:foos").should.eventually.eql("bar1");
+  });
+  it('should be able to use multiple find modules', async function () {
+    const d = new BaseDriver();
+    d.opts.customFindModules = {f: CUSTOM_FIND_MODULE, g: CUSTOM_FIND_MODULE};
+    await d.findElement(CUSTOM_STRATEGY, "f:foo").should.eventually.eql("bar");
```
KazuCocoa: 👍

jlipps:
thanks everyone! published as 3.9.0.

once appium 2 rolls around we can merge this into whatever the ultimate plugin structure becomes.

gh0st2sh3ll:
Hey @jlipps , is adding the custom locator strategy in the ruby-client on the horizon? :)

KazuCocoa:
let's try out the latest ruby_lib_core (2.0.4)
https://github.com/appium/ruby_lib_core/pull/151

SoundaryaBhavanam:
I am able to launch the app with the below capabilities
 HashMap<String, String> customFindModules = new HashMap<>();
 customFindModules.put("ai", "test-ai-classifier");
caps.setCapability("customFindModules", customFindModules);
 caps.setCapability("shouldUseCompactResponses", false);


But not able to find the text box with the visible text like 

driver.findElement(MobileBy.custom("ai:User ID"))

getting the below exception 

Element info: {Using=-custom, value=ai:User ID}
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at org.openqa.selenium.remote.ErrorHandler.createThrowable(ErrorHandler.java:214)
	at org.openqa.selenium.remote.ErrorHandler.throwIfResponseFailed(ErrorHandler.java:166)
	at org.openqa.selenium.remote.http.JsonHttpResponseCodec.reconstructValue(JsonHttpResponseCodec.java:40)
	at org.openqa.selenium.remote.http.AbstractHttpResponseCodec.decode(AbstractHttpResponseCodec.java:80)
	at org.openqa.selenium.remote.http.AbstractHttpResponseCodec.decode(AbstractHttpResponseCodec.java:44)
	at org.openqa.selenium.remote.HttpCommandExecutor.execute(HttpCommandExecutor.java:158)
	at io.appium.java_client.remote.AppiumCommandExecutor.execute(AppiumCommandExecutor.java:239)
	at org.openqa.selenium.remote.RemoteWebDriver.execute(RemoteWebDriver.java:548)
	at io.appium.java_client.DefaultGenericMobileDriver.execute(DefaultGenericMobileDriver.java:42)
	at io.appium.java_client.AppiumDriver.execute(AppiumDriver.java:1)
	at io.appium.java_client.ios.IOSDriver.execute(IOSDriver.java:1)
	at org.openqa.selenium.remote.RemoteWebDriver.findElement(RemoteWebDriver.java:322)
	at io.appium.java_client.DefaultGenericMobileDriver.findElement(DefaultGenericMobileDriver.java:62)
	at io.appium.java_client.AppiumDriver.findElement(AppiumDriver.java:1)
	at io.appium.java_client.ios.IOSDriver.findElement(IOSDriver.java:1)
	at io.appium.java_client.FindsByCustom.findElementByCustom(FindsByCustom.java:38)
	at io.appium.java_client.MobileBy$ByCustom.findElement(MobileBy.java:628)
	at org.openqa.selenium.remote.RemoteWebDriver.findElement(RemoteWebDriver.java:314)
	at io.appium.java_client.DefaultGenericMobileDriver.findElement(DefaultGenericMobileDriver.java:58)
	at io.appium.java_client.AppiumDriver.findElement(AppiumDriver.java:1)
	at io.appium.java_client.ios.IOSDriver.findElement(IOSDriver.java:1)
	at Testing.AI.appiumclassifierplugin.AppTest.testFindElementUsingAI(AppTest.java:90)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:124)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:580)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:716)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:988)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:125)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:109)
	at org.testng.TestRunner.privateRun(TestRunner.java:648)
	at org.testng.TestRunner.run(TestRunner.java:505)
	at org.testng.SuiteRunner.runTest(SuiteRunner.java:455)
	at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:450)
	at org.testng.SuiteRunner.privateRun(SuiteRunner.java:415)
	at org.testng.SuiteRunner.run(SuiteRunner.java:364)
	at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
	at org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:84)
	at org.testng.TestNG.runSuitesSequentially(TestNG.java:1208)
	at org.testng.TestNG.runSuitesLocally(TestNG.java:1137)
	at org.testng.TestNG.runSuites(TestNG.java:1049)
	at org.testng.TestNG.run(TestNG.java:1017)
	at org.testng.remote.AbstractRemoteTestNG.run(AbstractRemoteTestNG.java:114)
	at org.testng.remote.RemoteTestNG.initAndRun(RemoteTestNG.java:251)
	at org.testng.remote.RemoteTestNG.main(RemoteTestNG.java:77)
Can some one help me to find the elements ?

KazuCocoa:
@SoundaryaBhavanam 
Could you create the issue in https://github.com/testdotai/appium-classifier-plugin if your question is the AI logic?

PR #282, opened 2018-11-14
allow correct downloading of zipped apps without content type header

jlipps:
this case was missed in the refactor, causing a regression. should definitely get something like this in before 1.10

### DIFF
--- a/lib/basedriver/helpers.js
+++ b/lib/basedriver/helpers.js
@@ -94,13 +94,20 @@ async function configureApp (app, supportedAppExtensions) {
         APPLICATIONS_CACHE.del(app);
       }
       let fileName = `appium-app${_.first(supportedAppExtensions)}`;
+
+      // to determine if we need to unzip the app, we have a number of places
+      // to look: content type, content disposition, or the file extension
+      const downloadZipName = 'appium-app.zip';
       if (headers['content-type']) {
         logger.debug(`Content-Type: ${headers['content-type']}`);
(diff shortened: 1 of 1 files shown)


lib/basedriver/helpers.js
```diff
@@ -94,13 +94,20 @@ async function configureApp (app, supportedAppExtensions) {
...
+      const downloadZipName = 'appium-app.zip';
       if (headers['content-type']) {
         logger.debug(`Content-Type: ${headers['content-type']}`);
         // the filetype may not be obvious for certain urls, so check the mime type too
         if (ZIP_MIME_TYPES.includes(headers['content-type'])) {
-          fileName = 'appium-app.zip';
+          fileName = downloadZipName;
           shouldUnzipApp = true;
         }
+      } else if (ZIP_EXTS.includes(path.extname(newApp))) {
```
mykola-mokhnach: not sure if it should be `else if` or just `if`

jlipps: i was thinking we'd prefer content-type if it exists

mykola-mokhnach: it must be anyway very broken server, which does not return neither content type nor disposition for downloadable items

jlipps: that is github's server :-) or more likely what happens in the case of a redirect?

jlipps:
published as 3.11.2

PR #285, opened 2018-11-21
make allowing CORS conditional

jlipps:
It turns out blindly allowing CORS could be a security hazard (who knew)
See https://github.com/w3c/webdriver/pull/1362 for the conversation that led to this realization.
This change makes the cross domain middleware conditional (defaulting to the current behavior).
This way Appium can pass in 'false' based on a flag, for better server security.


jlipps:
related Appium PR coming

PR #296, opened 2019-01-08
add user-agent to list of cors headers

jlipps:
webdriverio apparently now sends the `user-agent` header, so for it to work over the web we need to add that to the cors header list.


jlipps:
cc @christian-bromann

imurchie:
I think it _is_ a good idea, in terms of debugging. Also metrics, if we could get a way to track such things.

jlipps:
published in 3.13.2

PR #297, opened 2019-01-26
Handle setValue API call conversion between protocols

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/12085

### DIFF
--- a/lib/jsonwp-proxy/protocol-converter.js
+++ b/lib/jsonwp-proxy/protocol-converter.js
@@ -129,6 +129,30 @@ class ProtocolConverter {
     return await this.proxyFunc(url, method, body);
   }
 
+  async proxySetValue (url, method, body) {
+    const bodyObj = util.safeJsonParse(body);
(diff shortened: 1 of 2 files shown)


KazuCocoa:
I found `org.json.JSONException` happened after this change. (Haven't tested iOS)

https://gist.github.com/KazuCocoa/39deaf62f134714812ac64bb4a0b954d

KazuCocoa:
Should we send text and value in webcontext case like iOS?

https://github.com/appium/appium-xcuitest-driver/blob/fcbbddc101fcfca23bad952dd104bd86dde8bd99/lib/commands/element.js#L163

https://github.com/appium/appium-uiautomator2-driver/blob/a91a34141634ecdb6928ecbd89312ef3042f1aa5/lib/commands/element.js#L60-L71

(Espresso driver handles `/value` command in the server)

KazuCocoa:
Do you have a time to take a look ^?
@mykola-mokhnach

PR #303, opened 2019-02-16
Fix application type handling for urls with parameters

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/12173

### DIFF
--- a/lib/basedriver/helpers.js
+++ b/lib/basedriver/helpers.js
@@ -22,6 +22,7 @@ const APPLICATIONS_CACHE = new LRU({
 });
 const APPLICATIONS_CACHE_GUARD = new AsyncLock();
 const SANITIZE_REPLACEMENT = '-';
+const DEFAULT_BASENAME = 'appium-app';
 
 async function retrieveHeaders (link) {
   try {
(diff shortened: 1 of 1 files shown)


lib/basedriver/helpers.js
```diff
@@ -96,37 +97,44 @@ async function configureApp (app, supportedAppExtensions) {
...
       if (!fileName) {
-        // assign the default extension if none has been detected
-        fileName = `appium-app${_.first(supportedAppExtensions)}`;
+        // assign the default file name and the extension if none has been detected
+        fileName = basename
+          ? basename.substring(0, basename.length - extname.length)
+          : DEFAULT_BASENAME;
+        fileName += supportedAppExtensions.includes(extname)
+          ? extname
+          : _.first(supportedAppExtensions);
```
dpgraham: I think it would be more readable if it was something like:

```javascript
basename = basename
          ? basename.substring(0, basename.length - extname.length)
          : DEFAULT_BASENAME;

extname = supportedAppExtensions.includes(extname)
          ? extname
          : _.first(supportedAppExtensions);

fileName = `${basename}${extname}`;
```

lib/basedriver/helpers.js
```diff
@@ -96,37 +97,44 @@ async function configureApp (app, supportedAppExtensions) {
...
         }
       }
       if (!fileName) {
-        // assign the default extension if none has been detected
-        fileName = `appium-app${_.first(supportedAppExtensions)}`;
+        // assign the default file name and the extension if none has been detected
+        fileName = basename
+          ? basename.substring(0, basename.length - extname.length)
+          : DEFAULT_BASENAME;
+        fileName += supportedAppExtensions.includes(extname)
```
dpgraham: Does this mean that if the `extname` is not supported it will just default to the first supported extension?

If so, maybe we should put a log message saying something like `'${extName}' is not supported. Defaulting to ${_.first(supportedAppExtensions}`

lib/basedriver/helpers.js
mykola-mokhnach: changed

PR #306, opened 2019-02-23
Fix find element wrong coordinate

KazuCocoa:
closes https://github.com/appium/appium/issues/12209

@jlipps Can you give me a good example app for `if (screenAR === shotAR)` lines I can confirm?
I simply commented out for now. If we do not need this, I would like to remove them.

---

will update https://github.com/appium/appium/blob/master/docs/en/advanced-concepts/image-elements.md later

---

# problem

_find by image_ feature returns wrong coordinate in Android and iOS case.
test cases in below test section fails even it compares same image/element.

## For Android

In [here](https://github.com/appium/appium-base-driver/blob/4eab674f92acfb1dd958fd48c9bbc9399d0d4e74/lib/basedriver/commands/find.js#L344-L359), Appium scales screenshot to window size. Then, Appium uses the scaled image as an input for `compareImages`.

e.g.
The screenshot is `1080x1920`. The window size is `1080x1794`. OpenCV compares a template image with the screenshot which is scaled from `1080x1920` to `1080x1794`. Then Appium returns `1080x1794` based coordinations if it succeeds to find the template image. (But the returns coordinate should be `1080x1920` based one, since `find_element :accessibility_id, xxxx` returns rect based on `1080x1920` based size.

## For iOS

iOS returns `750 × 1334 pixels` screenshot evern the window size is `width=375, height=667`.
`compareImages` uses scaled screenshot from `750 × 1334 pixels` to `width=375, height=667`. The ratio is 2.
Users should coordinate a template image scale if they would like to use `750 × 1334 pixels` screenshot based image as the template image to fixt to `width=375, height=667`.

It means:
Cannot find proper image

```ruby
el = @@driver.find_element :accessibility_id, 'Buttons' # find an element labelled 'buttons'
@driver.save_element_screenshot el, 'test_ios_button.png' # Save the element image

image_element = @driver.find_element_by_image 'test_ios_button.png' # Find image with the above element
image_element # raise no element found error because the input image is based on `750 × 1334 pixels` size
```

Then, users should do:
```ruby
el = @@driver.find_element :accessibility_id, 'Buttons' # find an element labelled 'buttons'
@driver.save_element_screenshot el, 'test_ios_button.png' # Save the element image

# Need a process to scale `test_ios_button.png` down to fit to `width=375, height=667` size

image_element = @driver.find_element_by_image 'test_ios_button.png'
image_element # can find image
```

# Implementation
- Get rid of [here](https://github.com/appium/appium-base-driver/blob/4eab674f92acfb1dd958fd48c9bbc9399d0d4e74/lib/basedriver/commands/find.js#L344-L359)
    - @jlipps Do you have a good example for this implementation?
- Introduce `fixScaleTemplateImage` for iOS case to reduce user actions for an input image

# test
https://github.com/appium/ruby_lib_core/pull/193/files is an example to make sure this PR works.
In the test, I ensured the below case for Android and iOS:

1. Find an element
2. Get the element screenshot
3. Find an element by image using _step2_ 's partial image
4. Assert _rect_ etc between `step1` and `step3`

They worked after this PR.
(This includes landscape and portrait mode.)

### DIFF
--- a/lib/basedriver/commands/find.js
+++ b/lib/basedriver/commands/find.js
@@ -386,7 +386,7 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
              `screen at ${screenWidth}x${screenHeight}`);
     imgObj = imgObj.resize(screenWidth, screenHeight);
 
-    scale = {xScale: screenWidth / shotWidth, yScale: screenHeight / shotHeight};
+    scale = {xScale: (1.0 * screenWidth) / shotWidth, yScale: (1.0 * screenHeight) / shotHeight};
   }
 
   b64Screenshot = (await imgObj.getBuffer(imageUtil.MIME_PNG)).toString('base64');
@@ -404,7 +404,7 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
  * @returns {string} base64-encoded scaled template screenshot
  */
 helpers.fixTemplateImageScale = async function (b64Template, scale) {
-  if (!scale) {
+  if (!scale || !_.isNumber(scale)) {
     return b64Template;
   }
 
--- a/lib/basedriver/commands/find.js
+++ b/lib/basedriver/commands/find.js
@@ -222,19 +222,9 @@ helpers.findByImage = async function (b64Template, {
   let rect = null;
   const condition = async () => {
     try {
-      let {b64Screenshot, recize} = await this.getScreenshotForImageFind(screenWidth, screenHeight);
+      const {b64Screenshot, scale} = await this.getScreenshotForImageFind(screenWidth, screenHeight);
 
-      if (recize) {
-        // For iOS case, mainly
-        let imgTempObj = await imageUtil.getJimpImage(b64Template);
-        let {width: baseTmpWidth, height: baseTmpHeigh} = imgTempObj.bitmap;
-        log.info(`Scaling template image from ${baseTmpWidth}x${baseTmpHeigh}` +
-                 ` to ${baseTmpWidth * recize.widthRatio}x${baseTmpHeigh * recize.heightRatio}`);
-        log.info(`The ratio is ${recize.widthRatio} and ${recize.heightRatio}`);
-        imgTempObj = await imgTempObj.resize(baseTmpWidth * recize.widthRatio, baseTmpHeigh * recize.heightRatio);
-        b64Template = (await imgTempObj.getBuffer(imageUtil.MIME_PNG)).toString('base64');
-      }
-
+      b64Template = await this.fixScaleTemplateImage(b64Template, scale);
       rect = (await this.compareImages(MATCH_TEMPLATE_MODE, b64Screenshot,
         b64Template, {threshold})).rect;
       return true;
@@ -314,13 +304,22 @@ helpers.ensureTemplateSize = async function (b64Template, screenWidth, screenHei
 };
 
 /**
+ * @typedef {Object} Screenshot
+ * @property {string} b64Screenshot - base64 based screenshot string
+ */
+/**
+ * @typedef {Object} ScreenshotScale
+ * @property {float} widthRatio - Scale ratio for width
+ * @property {float} heightRatio - Scale ratio for height
+ */
+/**
  * Get the screenshot image that will be used for find by element, potentially
  * altering it in various ways based on user-requested settings
  *
  * @param {int} screenWidth - width of screen
  * @param {int} screenHeight - height of screen
  *
- * @returns {string} base64-encoded screenshot
+ * @returns {Screenshot, ?ScreenshotScale} base64-encoded screenshot and ScreenshotScale
  */
 helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
   if (!this.getScreenshot) {
@@ -333,7 +332,7 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
   // between the screenshot and the screen, just return the screenshot now
   if (!this.settings.getSettings().fixImageFindScreenshotDims) {
     log.info(`Not verifying screenshot dimensions match screen`);
-    return b64Screenshot;
+    return {b64Screenshot};
   }
 
   // otherwise, do some verification on the screenshot to make sure it matches
@@ -347,7 +346,7 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
     // the height and width of the screenshot and the device screen match, which
     // means we should be safe when doing template matches
     log.info('Screenshot size matched screen size');
-    return b64Screenshot;
+    return {b64Screenshot};
   }
 
   // otherwise, if they don't match, it could spell problems for the accuracy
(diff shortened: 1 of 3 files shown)


lib/basedriver/commands/find.js
```diff
@@ -341,35 +359,74 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
-    log.warn(`Resizing screenshot to ${shotWidth}x${shotHeight} to match ` +
-             `screen aspect ratio so that image element coordinates have a ` +
-             `greater chance of being correct.`);
-    imgObj = imgObj.resize(shotWidth, shotHeight);
-  }
-
-  // now we know the aspect ratios match, but there might still be a scale
-  // mismatch, so just resize based on the screen dimensions
-  if (screenWidth !== shotWidth) {
+  // const screenAR = screenWidth / screenHeight;
```
KazuCocoa: Will remove here if we do not need these lines

lib/basedriver/commands/find.js
```diff
@@ -201,7 +201,8 @@ helpers.findByImage = async function (b64Template, {
 }) {
   const {
     imageMatchThreshold: threshold,
-    fixImageTemplateSize
+    fixImageTemplateSize,
+    fixScaleTemplateImage
```
mykola-mokhnach: perhaps `fixTemplateImageScale` ?

lib/basedriver/commands/find.js
```diff
@@ -288,24 +294,36 @@ helpers.ensureTemplateSize = async function (b64Template, screenWidth, screenHei
...
   return (await imgObj.getBuffer(imageUtil.MIME_PNG)).toString('base64');
 };
 
+/**
+ * @typedef {Object} Screenshot
+ * @property {string} b64Screenshot - base64 based screenshot string
+ */
+/**
+ * @typedef {Object} ScreenshotScale
+ * @property {float} widthRatio - Scale ratio for width
```
mykola-mokhnach: xScale, yScale

lib/basedriver/commands/find.js
```diff
@@ -222,7 +223,12 @@ helpers.findByImage = async function (b64Template, {
   let rect = null;
   const condition = async () => {
     try {
-      let b64Screenshot = await this.getScreenshotForImageFind(screenWidth, screenHeight);
+      const {b64Screenshot, scale} = await this.getScreenshotForImageFind(screenWidth, screenHeight);
+
+      if (fixScaleTemplateImage) {
```
mykola-mokhnach: should we also check if `scale` is defined?

lib/basedriver/commands/find.js
```diff
@@ -341,35 +359,74 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
+  // since except for that, it might be a situation which is different window rect and
+  // screenshot size like `@driver.window_rect #=>x=0, y=0, width=1080, height=1794` and
+  // `"deviceScreenSize"=>"1080x1920"`
+  let scale;
+  if (screenWidth !== shotWidth && screenHeight !== shotHeight) {
     log.info(`Scaling screenshot from ${shotWidth}x${shotHeight} to match ` +
              `screen at ${screenWidth}x${screenHeight}`);
     imgObj = imgObj.resize(screenWidth, screenHeight);
+
+    scale = {widthRatio: screenWidth / shotWidth, heightRatio: screenHeight / shotHeight};
```
mykola-mokhnach: this might be an integer division. how about adding `1.0 *` to each expression?

lib/basedriver/commands/find.js
```diff
@@ -341,35 +359,74 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
   }
 
-  return (await imgObj.getBuffer(imageUtil.MIME_PNG)).toString('base64');
+  b64Screenshot = (await imgObj.getBuffer(imageUtil.MIME_PNG)).toString('base64');
+  return {b64Screenshot, scale};
 };
 
 
+/**
```
mykola-mokhnach: Why do we define it twice?

lib/basedriver/commands/find.js
```diff
@@ -341,35 +359,74 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
+ * @property {float} heightRatio - Scale ratio for height
+ */
+
+/**
+ * Get a image that will be used for template maching.
+ * Returns scaled image if scale ratio is provided.
+ *
+ * @param {string} b64Template - base64-encoded image used as a template to be
+ * matched in the screenshot
+ * @param {?ScreenshotScale} scale - height of screen
```
mykola-mokhnach: height?

lib/basedriver/commands/find.js
```diff
@@ -341,35 +359,74 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
+ *
+ * @returns {string} base64-encoded scaled template screenshot
+ */
+helpers.fixScaleTemplateImage = async function (b64Template, scale) {
+  if (!scale) {
+    return b64Template;
+  }
```
mykola-mokhnach: `scale` might be destructed here

lib/basedriver/commands/find.js
KazuCocoa: mm. If the result is a float, then will this float?

```node
> 1098 / 500
2.196
```

I quick checked. `1.0` work as an integer in JS??

```node
> function isFloat(n){
...     return Number(n) === n && n % 1 !== 0;
... }
undefined
> a = 1.1
1.1
> isFloat(a)
true
> a = 1.0
1
> isFloat(a)
false
> parseFloat(a)
1
> isFloat(parseFloat(a))
false
```

lib/basedriver/commands/find.js
```diff
@@ -201,7 +201,8 @@ helpers.findByImage = async function (b64Template, {
 }) {
   const {
     imageMatchThreshold: threshold,
-    fixImageTemplateSize
+    fixImageTemplateSize,
+    fixTemplateImageScale
```
mykola-mokhnach: lets be consistent fixImageTemplateSize + fixImageTemplateScale %)

lib/basedriver/commands/find.js
mykola-mokhnach: yep, I vote for removing the commented stuff

lib/basedriver/commands/find.js
```diff
@@ -341,34 +359,66 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
+  return {b64Screenshot, scale};
 };
 
+/**
+ * Get a image that will be used for template maching.
+ * Returns scaled image if scale ratio is provided.
+ *
+ * @param {string} b64Template - base64-encoded image used as a template to be
+ * matched in the screenshot
+ * @param {?ScreenshotScale} scale - scale of screen
```
mykola-mokhnach: it might be better for the readability if you provide the default value for the argument if it is optional (`scale=null`)

lib/basedriver/commands/find.js
mykola-mokhnach: I mean `const {xScale, yScale} = scale`

lib/basedriver/commands/find.js
mykola-mokhnach: yeah, I'm constantly forgetting that Javascript has many unordinary things like the missing integer division concept, for example :)

KazuCocoa: ah.. gotha :)

KazuCocoa:
I'm about to merge this in order to publish this via at beta.
I'll re-consider https://github.com/appium/appium-base-driver/pull/306/commits/4fb91b9fd335e333b36586f51164e70a37b18aea again if the beta has ratio issue.
(Perhaps, scaling width or hight base is pretty enough in many cases.)

lib/basedriver/commands/find.js
```diff
@@ -332,43 +350,63 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
-  // mismatch, so just resize based on the screen dimensions
-  if (screenWidth !== shotWidth) {
+  // mismatch.
+
+  // Resize based on the screen dimensions only if both width and height are mismatched
+  // since except for that, it might be a situation which is different window rect and
+  // screenshot size like `@driver.window_rect #=>x=0, y=0, width=1080, height=1794` and
+  // `"deviceScreenSize"=>"1080x1920"`
+  let scale;
+  if (screenWidth !== shotWidth && screenHeight !== shotHeight) {
```
KazuCocoa: Will consider different ratio case more (and create a PR)

PR #307, opened 2019-03-01
Fix template scale in find element by image

KazuCocoa:
Fix remaining issue in https://github.com/appium/appium-base-driver/pull/306/files#r261480026 .
In the PR, I fixed a template image had different width/height with device screen, but it can be fixed scaling them up/down upto match to device width/height.

But the ratio issue has also below case.

example:

- A user has a template image which is scaled down from `750 x 1314 pixels` image to `187 x 328 pixels` in order to decrease their storage when he store the template image.
- The user uses the scale downed image as the next test case
    - Then, the scaled image does not match in current find by image logic since Appium cannot determin what ratio should be scaled the template up/down
    - Users must scale them up the image in the user side every time.

To resolve the case, I introduced `defaultImageTemplateScale`. If the user sets `defaultImageTemplateScale: 4` for example, Appium will scale the template image up 4 times by default. Then, users do not need to scale in user side.

It can use with `fixImageTemplateScale` for example. It can help to handle some cases such as I described as https://github.com/appium/appium-base-driver/pull/306

cc @jlipps

### DIFF
--- a/lib/basedriver/commands/find.js
+++ b/lib/basedriver/commands/find.js
@@ -12,6 +12,10 @@ const commands = {}, helpers = {}, extensions = {};
 const IMAGE_STRATEGY = '-image';
 const CUSTOM_STRATEGY = '-custom';
 
+// Used to compar ratio and screen width
+// Pixel is basically under 1080 for example. 100K is probably enough fo a while.
+const DEFAULT_ROUND_FLOAT = 100000;
+
 // Override the following function for your own driver, and the rest is taken
 // care of!
 
@@ -369,7 +373,7 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
 
   const screenAR = screenWidth / screenHeight;
   const shotAR = shotWidth / shotHeight;
-  if (screenAR === shotAR) {
+  if (Math.round(screenAR * DEFAULT_ROUND_FLOAT) === Math.round(shotAR * DEFAULT_ROUND_FLOAT)) {
     log.info('Screenshot aspect ratio matched screen aspect ratio');
   } else {
     log.warn(`When trying to find an element, determined that the screen ` +
@@ -473,8 +477,9 @@ helpers.fixImageTemplateScale = async function (b64Template, opts = {}) {
     return b64Template;
   }
 
-  // Return if the scale is default (1.0) value
-  if ((xScale - DEFAULT_FIX_IMAGE_TEMPLATE_SCALE) <= 0 && (yScale - DEFAULT_FIX_IMAGE_TEMPLATE_SCALE) === 0) {
+  // Return if the scale is default, 1, value
+  if (Math.round(xScale * DEFAULT_ROUND_FLOAT) === Math.round(DEFAULT_FIX_IMAGE_TEMPLATE_SCALE * DEFAULT_ROUND_FLOAT)
+      && Math.round(yScale * DEFAULT_ROUND_FLOAT === Math.round(DEFAULT_FIX_IMAGE_TEMPLATE_SCALE * DEFAULT_ROUND_FLOAT))) {
     return b64Template;
   }
 
--- a/lib/basedriver/commands/find.js
+++ b/lib/basedriver/commands/find.js
@@ -203,7 +203,7 @@ helpers.findByImage = async function (b64Template, {
     imageMatchThreshold: threshold,
     fixImageTemplateSize,
     fixImageTemplateScale,
-    defaultScaleImageTemplate
+    defaultImageTemplateScale
   } = this.settings.getSettings();
 
   log.info(`Finding image element with match threshold ${threshold}`);
@@ -227,7 +227,7 @@ helpers.findByImage = async function (b64Template, {
       const {b64Screenshot, scale} = await this.getScreenshotForImageFind(screenWidth, screenHeight);
 
       b64Template = await this.fixImageTemplateScale(b64Template, {
-        defaultScaleImageTemplate, fixImageTemplateScale, ...scale
+        defaultImageTemplateScale, fixImageTemplateScale, ...scale
       });
 
       rect = (await this.compareImages(MATCH_TEMPLATE_MODE, b64Screenshot,
@@ -380,7 +380,7 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
 /**
  * @typedef {Object} ImageTemplateSettings
  * @property {float} fixImageTemplateScale - fixImageTemplateScale in device-settings
- * @property {float} defaultScaleImageTemplate - defaultScaleImageTemplate in device-settings
+ * @property {float} defaultImageTemplateScale - defaultImageTemplateScale in device-settings
  */
 /**
  * Get a image that will be used for template maching.
@@ -398,35 +398,35 @@ helpers.fixImageTemplateScale = async function (b64Template, opts = {}) {
     return b64Template;
   }
 
-  const default_scale = 1.0;
-  const defaultSettings = {
+  const defaultScale = 1.0;
+  const defaultOpts = {
     fixImageTemplateScale: false,
-    defaultScaleImageTemplate: DEFAULT_TEMPLATE_IMAGE_SCALE,
-    xScale: default_scale,
-    yScale: default_scale
+    defaultImageTemplateScale: DEFAULT_TEMPLATE_IMAGE_SCALE,
+    xScale: defaultScale,
+    yScale: defaultScale
   };
-  _.defaults(opts, defaultSettings);
+  _.defaults(opts, defaultOpts);
 
   let {xScale, yScale} = opts;
   // Default
-  if (opts.defaultScaleImageTemplate === DEFAULT_TEMPLATE_IMAGE_SCALE && !opts.fixImageTemplateScale) {
+  if (opts.defaultImageTemplateScale === DEFAULT_TEMPLATE_IMAGE_SCALE && !opts.fixImageTemplateScale) {
     return b64Template;
   }
 
   if (opts.fixImageTemplateScale) {
-    xScale = xScale * opts.defaultScaleImageTemplate;
-    yScale = yScale * opts.defaultScaleImageTemplate;
+    xScale = xScale * opts.defaultImageTemplateScale;
+    yScale = yScale * opts.defaultImageTemplateScale;
   } else {
-    xScale = yScale = 1.0 * opts.defaultScaleImageTemplate;
+    xScale = yScale = 1.0 * opts.defaultImageTemplateScale;
   }
 
-  // xScale and yScale can be NaN if opts.defaultScaleImageTemplate is string, for example
+  // xScale and yScale can be NaN if opts.defaultImageTemplateScale is string, for example
   if (!xScale || !yScale) {
     return b64Template;
   }
 
   // Return if the scale is deatul (1.0) value
-  if (xScale === default_scale && yScale === default_scale) {
+  if (xScale === defaultScale && yScale === defaultScale) {
     return b64Template;
   }
 
--- a/lib/basedriver/commands/find.js
+++ b/lib/basedriver/commands/find.js
@@ -365,7 +365,7 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
   // are two potential types of mismatch: aspect ratio mismatch and scale
   // mismatch. we need to detect and fix both
 
-  let scale = {xScale: 1.0, yScale: 1.0};
+  const scale = {xScale: 1.0, yScale: 1.0};
 
   const screenAR = screenWidth / screenHeight;
   const shotAR = shotWidth / shotHeight;
--- a/lib/basedriver/commands/find.js
+++ b/lib/basedriver/commands/find.js
@@ -400,6 +400,9 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
 
     scale.xScale *= scaleFactor;
     scale.yScale *= scaleFactor;
+
+    shotWidth = imgObj.bitmap.width;
+    shotHeight = imgObj.bitmap.height;
   }
 
   // Resize based on the screen dimensions only if both width and height are mismatched
--- a/lib/basedriver/commands/find.js
+++ b/lib/basedriver/commands/find.js
@@ -2,7 +2,7 @@ import log from '../logger';
 import { logger, imageUtil } from 'appium-support';
 import _ from 'lodash';
 import { errors } from '../../..';
-import { MATCH_TEMPLATE_MODE } from './images';
+import { MATCH_TEMPLATE_MODE, DEFAULT_TEMPLATE_IMAGE_SCALE } from './images';
 import { W3C_ELEMENT_KEY, MJSONWP_ELEMENT_KEY } from '../../protocol/protocol';
 import { ImageElement } from '../image-element';
 
@@ -203,7 +203,7 @@ helpers.findByImage = async function (b64Template, {
     imageMatchThreshold: threshold,
     fixImageTemplateSize,
     fixImageTemplateScale,
-    imageTemplateDefaultScale
+    defaultScaleImageTemplate
   } = this.settings.getSettings();
 
   log.info(`Finding image element with match threshold ${threshold}`);
@@ -226,13 +226,9 @@ helpers.findByImage = async function (b64Template, {
     try {
       const {b64Screenshot, scale} = await this.getScreenshotForImageFind(screenWidth, screenHeight);
 
-      if (imageTemplateDefaultScale) {
-        b64Template = await this.setImageTemplateDefaultScale(b64Template, imageTemplateDefaultScale)
-      }
-
-      if (fixImageTemplateScale) {
-        b64Template = await this.fixImageTemplateScale(b64Template, scale);
-      }
+      b64Template = await this.fixImageTemplateScale(b64Template, {
+        defaultScaleImageTemplate, fixImageTemplateScale, ...scale
+      });
 
       rect = (await this.compareImages(MATCH_TEMPLATE_MODE, b64Screenshot,
         b64Template, {threshold})).rect;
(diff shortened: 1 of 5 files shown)


lib/basedriver/commands/find.js
```diff
@@ -183,6 +183,9 @@ commands.findByCustom = async function (selector, multiple) {
  * image is merely to check staleness. If so we can bypass a lot of logic
  * @property {boolean} [multiple=false] - Whether we are finding one element or
  * multiple
+ * @property {boolean} [ignoreDefaultImageTemplateScale=false] - Whether we
+ * ignore defaultImageTemplateScale. It can be used when you would noe like to
```
mykola-mokhnach: noe?

lib/basedriver/commands/find.js
```diff
@@ -376,26 +382,72 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
   return {b64Screenshot, scale};
 };
 
+/**
+ * @typedef {Object} ImageTemplateSettings
+ * @property {boolean} fixImageTemplateScale - fixImageTemplateScale in device-settings
+ * @property {float} defaultImageTemplateScale - defaultImageTemplateScale in device-settings
+ * @property {boolean} ignoreDefaultImageTemplateScale - Ignore defaultImageTemplateScale if it has true.
```
mykola-mokhnach: please include `scale` to this typedef

lib/basedriver/commands/find.js
```diff
@@ -376,26 +382,72 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
+ */
 /**
  * Get a image that will be used for template maching.
  * Returns scaled image if scale ratio is provided.
  *
+ *
  * @param {string} b64Template - base64-encoded image used as a template to be
  * matched in the screenshot
- * @param {ScreenshotScale} scale - scale of screen. Default to `{}`
+ * @param {ScreenshotScale, ImageTemplateSettings} opts - Image template scale related options
```
mykola-mokhnach: I'd rather keep a single ImageTemplateSettings type here.

lib/basedriver/commands/find.js
```diff
@@ -376,26 +382,72 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
-helpers.fixImageTemplateScale = async function (b64Template, scale = {}) {
-  if (!scale) {
+helpers.fixImageTemplateScale = async function (b64Template, opts = {}) {
+  if (!opts) {
+    return b64Template;
+  }
+
+  const defaultScale = 1.0; // To initialise xScale and yScale
+  const defaultOpts = {
+    fixImageTemplateScale: false,
```
mykola-mokhnach: You can set defaults without _.defaults call:

```
const {
  defaultImageTemplateScale: DEFAULT_TEMPLATE_IMAGE_SCALE,
  ...
} = opts;

```

lib/basedriver/commands/find.js
```diff
@@ -376,26 +382,72 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
  * @returns {string} base64-encoded scaled template screenshot
  */
-helpers.fixImageTemplateScale = async function (b64Template, scale = {}) {
-  if (!scale) {
+helpers.fixImageTemplateScale = async function (b64Template, opts = {}) {
+  if (!opts) {
+    return b64Template;
+  }
+
+  const defaultScale = 1.0; // To initialise xScale and yScale
```
mykola-mokhnach: constants should be in uppercase

mykola-mokhnach: also it's better to move it out of the func

lib/basedriver/commands/find.js
```diff
@@ -376,26 +382,72 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
     return b64Template;
   }
 
-  const {xScale, yScale} = scale;
+  // Calculate xScale and yScale Appium should scale
+  if (opts.fixImageTemplateScale) {
+    xScale = xScale * defaultImageTemplateScale;
+    yScale = yScale * defaultImageTemplateScale;
+  } else {
+    // `1.0 *` makes NaN if defaultImageTemplateScale is no number
```
mykola-mokhnach: is not a number

lib/basedriver/commands/find.js
```diff
@@ -376,26 +382,72 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
+  let {xScale, yScale} = opts;
+  // Default
+  if (defaultImageTemplateScale === DEFAULT_TEMPLATE_IMAGE_SCALE && !opts.fixImageTemplateScale) {
     return b64Template;
   }
 
-  const {xScale, yScale} = scale;
+  // Calculate xScale and yScale Appium should scale
+  if (opts.fixImageTemplateScale) {
+    xScale = xScale * defaultImageTemplateScale;
```
mykola-mokhnach: use xScale *= defaultImageTemplateScale

lib/basedriver/commands/find.js
```diff
@@ -376,26 +382,72 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
+    // `1.0 *` makes NaN if defaultImageTemplateScale is no number
+    xScale = yScale = 1.0 * defaultImageTemplateScale;
+  }
+
+  // xScale and yScale can be NaN if opts.defaultImageTemplateScale is string, for example
   if (!xScale || !yScale) {
     return b64Template;
   }
 
+  // Return if the scale is deatul (1.0) value
```
mykola-mokhnach: deatul?

lib/basedriver/commands/find.js
```diff
@@ -376,26 +382,72 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
+  if (xScale === defaultScale && yScale === defaultScale) {
```
mykola-mokhnach: in general it is not a good idea to compare float values for equality

lib/basedriver/device-settings.js
```diff
@@ -25,6 +25,14 @@ const GLOBAL_DEFAULT_SETTINGS = {
   //      if a user use `width=750 × height=1334` pixels's base template image.
   fixImageTemplateScale: false,
 
+  // Users might have scaled template image to reduce their storage size.
```
mykola-mokhnach: I would be nice to have this also described in some doc

KazuCocoa: Yes. I'm thinking to add this in https://github.com/appium/appium/blob/master/docs/en/advanced-concepts/image-elements.md

lib/basedriver/commands/find.js
```diff
@@ -357,45 +363,120 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
   // on the screenshot coordinates not the device coordinates themselves. There
   // are two potential types of mismatch: aspect ratio mismatch and scale
-  // mismatch.
+  // mismatch. we need to detect and fix both
+
+  let scale = {xScale: 1.0, yScale: 1.0};
+
+  const screenAR = screenWidth / screenHeight;
+  const shotAR = shotWidth / shotHeight;
+  if (screenAR === shotAR) {
```
KazuCocoa: https://github.com/appium/appium-base-driver/pull/306/files#diff-a7d1893e663367145d766d1aaab4aad4L347 case.
I discussed with @jlipps and figure out the issue.
I left an example of this ratio issue as a comment below.

mykola-mokhnach: float comparison?

lib/basedriver/commands/find.js
```diff
@@ -357,45 +363,122 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
+
+    // Select smaller ratio
+    // this.getScreenshot(shotWidth, shotHeight) is 540x397,
+    // this.getDeviceSize(screenWidth, screenHeight) is 1080x1920.
+    // The ratio is {xScale: 0.5, yScale: 0.2}.
+    // In this case, we must choose `yScale: 0.2` as scaleFactor.
+    // Because if Appium selects the both ratio, the screenshot will be distorted.
+    // If Appium selects the xScale, height will be bigger than real screenshot size
+    // which is used to image comparison by OpenCV as a base image.
+    const _xScale = (1.0 * shotWidth) / screenWidth;
```
mykola-mokhnach: do we need underscore prefixes?

lib/basedriver/commands/find.js
```diff
@@ -357,45 +363,122 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
-helpers.fixImageTemplateScale = async function (b64Template, scale = {}) {
-  if (!scale) {
+const DEFAULT_FIX_IMAGE_TEMPLATE_SCALE = 1;
+helpers.fixImageTemplateScale = async function (b64Template, opts = {}) {
+  if (!opts) {
     return b64Template;
   }
 
-  const {xScale, yScale} = scale;
+  let {
```
mykola-mokhnach: 👍

lib/basedriver/commands/find.js
```diff
@@ -357,45 +363,122 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
+  if (defaultImageTemplateScale === DEFAULT_TEMPLATE_IMAGE_SCALE && !fixImageTemplateScale) {
+    return b64Template;
+  }
+
+  // Calculate xScale and yScale Appium should scale
+  if (fixImageTemplateScale) {
+    xScale *= defaultImageTemplateScale;
+    yScale *= defaultImageTemplateScale;
+  } else {
+    // `1 *` makes NaN if defaultImageTemplateScale is not a number
```
mykola-mokhnach: I think parseFloat  would be more readable for such purpose

lib/basedriver/commands/find.js
```diff
@@ -357,45 +363,122 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
+    xScale = yScale = 1 * defaultImageTemplateScale;
+  }
+
+  // xScale and yScale can be NaN if defaultImageTemplateScale is string, for example
   if (!xScale || !yScale) {
     return b64Template;
   }
 
+  // Return if the scale is default (1.0) value
+  if ((xScale - DEFAULT_FIX_IMAGE_TEMPLATE_SCALE) <= 0 && (yScale - DEFAULT_FIX_IMAGE_TEMPLATE_SCALE) === 0) {
```
mykola-mokhnach: please avoid comparing float values for equality.
It is mandatory to have some delta epsilon value or use toFixed transformation and compare these are strings

mykola-mokhnach: https://stackoverflow.com/questions/3343623/javascript-comparing-two-float-values

lib/basedriver/commands/find.js
KazuCocoa:
```
> parseFloat('a')
NaN
```
ah... 👍

lib/basedriver/commands/find.js
```diff
@@ -12,6 +12,10 @@ const commands = {}, helpers = {}, extensions = {};
 const IMAGE_STRATEGY = '-image';
 const CUSTOM_STRATEGY = '-custom';
 
+// Used to compar ratio and screen width
+// Pixel is basically under 1080 for example. 100K is probably enough fo a while.
+const DEFAULT_ROUND_FLOAT = 100000;
```
mykola-mokhnach: I'd rather rename it to FLOAT_PRECISION

lib/basedriver/commands/find.js
```diff
@@ -4,14 +4,18 @@ import _ from 'lodash';
...
 
 
 const commands = {}, helpers = {}, extensions = {};
 
 const IMAGE_STRATEGY = '-image';
 const CUSTOM_STRATEGY = '-custom';
 
+// Used to compar ratio and screen width
+// Pixel is basically under 1080 for example. 100K is probably enough fo a while.
+const FLOAT_PERCISION = 100000;
```
mykola-mokhnach: PRECISION

lib/basedriver/commands/find.js
```diff
@@ -357,42 +367,119 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
   // of coordinates returned by the image match algorithm, since we match based
   // on the screenshot coordinates not the device coordinates themselves. There
   // are two potential types of mismatch: aspect ratio mismatch and scale
-  // mismatch.
+  // mismatch. we need to detect and fix both
```
mykola-mokhnach: We

lib/basedriver/commands/find.js
KazuCocoa: oops...

KazuCocoa:
Will wait for @jlipps to double check the comparison logic

KazuCocoa:
@jlipps Do you have a chance to have a look?
(Or can I merge and publish it as beta?)

lib/basedriver/commands/find.js
```diff
@@ -4,14 +4,18 @@ import _ from 'lodash';
...
-import { ImageElement } from '../image-element';
+import { ImageElement, DEFAULT_TEMPLATE_IMAGE_SCALE } from '../image-element';
 
 
 const commands = {}, helpers = {}, extensions = {};
 
 const IMAGE_STRATEGY = '-image';
 const CUSTOM_STRATEGY = '-custom';
 
+// Used to compar ratio and screen width
```
jlipps: compare

lib/basedriver/commands/find.js
```diff
@@ -357,42 +367,119 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
   // are two potential types of mismatch: aspect ratio mismatch and scale
-  // mismatch.
+  // mismatch. We need to detect and fix both
+
+  const scale = {xScale: 1.0, yScale: 1.0};
+
+  const screenAR = screenWidth / screenHeight;
+  const shotAR = shotWidth / shotHeight;
+  if (Math.round(screenAR * FLOAT_PRECISION) === Math.round(shotAR * FLOAT_PRECISION)) {
+    log.info('Screenshot aspect ratio matched screen aspect ratio');
```
jlipps: would be nice to do something like 'Screenshot aspect ratio (${shotAR}) matched screen aspect ratio (${screenAR})'

lib/basedriver/commands/find.js
```diff
@@ -357,42 +367,119 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
+  } else {
+    log.warn(`When trying to find an element, determined that the screen ` +
+             `aspect ratio and screenshot aspect ratio are different. Screen ` +
+             `is ${screenWidth}x${screenHeight} whereas screenshot is ` +
+             `${shotWidth}x${shotHeight}.`);
+
+    // Select smaller ratio
```
jlipps:
```
    // In the case where the x-scale and y-scale are different, we need to decide
    // which one to respect, otherwise the screenshot and template will end up
    // being resized in a way that changes its aspect ratio (distorts it). For example, let's say:
    // this.getScreenshot(shotWidth, shotHeight) is 540x397,
    // this.getDeviceSize(screenWidth, screenHeight) is 1080x1920.
    // The ratio would then be {xScale: 0.5, yScale: 0.2}.
    // In this case, we must should `yScale: 0.2` as scaleFactor, because
    // If we select the xScale, the height will be bigger than real screenshot size
    // which is used to image comparison by OpenCV as a base image.
    // All of this is primarily useful when the screenshot is a horizontal slice taken out of the
    // screen (for example not including top/bottom nav bars)
```

lib/basedriver/commands/find.js
```diff
@@ -357,42 +367,119 @@ helpers.getScreenshotForImageFind = async function (screenWidth, screenHeight) {
...
+    scale.xScale *= scaleFactor;
+    scale.yScale *= scaleFactor;
+  }
 
   // Resize based on the screen dimensions only if both width and height are mismatched
   // since except for that, it might be a situation which is different window rect and
   // screenshot size like `@driver.window_rect #=>x=0, y=0, width=1080, height=1794` and
   // `"deviceScreenSize"=>"1080x1920"`
-  let scale;
   if (screenWidth !== shotWidth && screenHeight !== shotHeight) {
```
jlipps: shouldn't we be comparing the _new_ imgObj width and height here? it looks like we're comparing against the old shotWidth/shotHeight, but those could have changed in the previous block, right?

lib/basedriver/device-settings.js
```diff
@@ -25,6 +25,14 @@ const GLOBAL_DEFAULT_SETTINGS = {
...
+  // This setting allows users to scale a template image they send to Appium server
+  // so that Appium server compare actual scale users orginaly have.
+  // e.g. A user have `270 × 32 pixels` image oridinally `1080 × 126 pixels`
```
jlipps: originally

lib/basedriver/device-settings.js
```diff
@@ -25,6 +25,14 @@ const GLOBAL_DEFAULT_SETTINGS = {
   //      if a user use `width=750 × height=1334` pixels's base template image.
   fixImageTemplateScale: false,
 
+  // Users might have scaled template image to reduce their storage size.
+  // This setting allows users to scale a template image they send to Appium server
+  // so that Appium server compare actual scale users orginaly have.
```
jlipps: so that the Appium server compares the actual scale users originally had

lib/basedriver/device-settings.js
jlipps: If a user has an image of 270 x 32 pixels which was originally 1080 x 126 pixels

lib/basedriver/device-settings.js
```diff
@@ -25,6 +25,14 @@ const GLOBAL_DEFAULT_SETTINGS = {
...
+  // e.g. A user have `270 × 32 pixels` image oridinally `1080 × 126 pixels`
+  //      The user set {defaultImageTemplateScale: 4.0} to scale the small image
```
jlipps: The user can set

lib/basedriver/commands/find.js
KazuCocoa: Right. I added https://github.com/appium/appium-base-driver/pull/307/commits/b536b1039766f45e95ee5be3c79d68a3f9e84304 to update `shotWidth` and `shotHeight` if they should be updated in the ratio comparison

KazuCocoa: @jlipps Do you have a chance to take a look ^?

PR #308, opened 2019-03-05
Fixed typo signalStrength

khanhdodang:


lib/protocol/routes.js
```diff
@@ -387,7 +387,7 @@ const METHOD_MAP = {
     POST: {command: 'gsmCall', payloadParams: {required: ['phoneNumber', 'action']}}
   },
   '/wd/hub/session/:sessionId/appium/device/gsm_signal': {
-    POST: {command: 'gsmSignal', payloadParams: {required: ['signalStrengh']}}
+    POST: {command: 'gsmSignal', payloadParams: {required: ['signalStrength']}}
```
KazuCocoa: As a backward compatibility, we can do like https://github.com/appium/appium-base-driver/blob/137a664979130020651f06de6c6a9aa9f9a06821/lib/protocol/routes.js#L180-L194

```
[jsonObj. signalStrength || jsonObj. signalStrengh]
```

and clients can send both of them
(can you work for it?)

khanhdodang: Hi @KazuCocoa This is a typo issue. I don't think we need to handle backward compatibility here. I also made these PRs to fix this issue on clients. Please see other thread from client https://github.com/webdriverio/webdriverio/pull/3678

imurchie: I think backward compatibility is necessary until _all_ the clients are changed to handle this.

KazuCocoa: Yes
I confirmed related links in this PRs. Java and wd clients also have it in additon to webdriverio and ruby client. (thanks for them)

It is easy to allow both of them for a while as @imurchie commented.
Clients also send both params. Then, they can be released separately without consideration by users. Some users use newer clients with older Appium. The opposite also happens.

khanhdodang: @KazuCocoa @imurchie I have updated the PR.

PR #310, opened 2019-03-22
Update bluebird

imurchie:
This is the last package to be using `bluebird@2`. We did not upgrade because of the change in the cancellable semantics: the effective handling of unexpected errors relies on cancelling promises _aborting_ them, while `bluebird@3` changed it to simply nullifying listeners.

This PR rolls a simple abort-cancel, so the functionality remains the same.

### DIFF
--- a/lib/basedriver/driver.js
+++ b/lib/basedriver/driver.js
@@ -295,12 +295,14 @@ class BaseDriver extends Protocol {
           const cancelPromise = new B(function (_, _reject) { // eslint-disable-line no-unused-vars
             reject = _reject;
           });
+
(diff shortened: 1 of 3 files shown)


mykola-mokhnach:
Some black magic there, but I'm ok with it if the solution works as expected

PeoB:
Hey, my automated tests started crashing at startup after this PR with ` Error: cannot enable cancellation after promises are in use`.

Any suggestion? Commenting out https://github.com/appium/appium-base-driver/pull/310/files#diff-69c68e3397852a86397892936887db6aR16 fixes the problem.

PR #311, opened 2019-03-28
Make proxy convert /property to /attribute

dpgraham:
After digging into https://github.com/appium/appium/issues/12218 I think the best solution is to proxy W3C `/property/:prop` commands through to JSONWP as `/attribute/:attr`.

So, for example, if a W3C session has this endpoint called

`GET: /session/123/element/456/property/value`

and it's being proxied to a JSONWP server, it should be proxied as

`GET: /session/123/element/456/attribute/value`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: node_js
 node_js:
   - "8"
   - "10"
-  - "11"
 addons:
   apt:
     sources:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 node_js:
   - "8"
   - "10"
+  - "11"
 addons:
   apt:
     sources:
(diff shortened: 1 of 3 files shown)


lib/jsonwp-proxy/protocol-converter.js
```diff
@@ -35,6 +35,16 @@ const COMMAND_URLS_CONFLICTS = [
         : url.replace(/\/window_handles$/, '/window/handles');
     },
   },
+  {
+    commandNames: ['getProperty'],
+    jsonwpConverter: (w3cUrl) => {
+      const w3cPropertyRegex = /\/element\/([^/]+)\/property\/([^/]+)/;
```
mykola-mokhnach: is trailing slash mandatory there?

mykola-mokhnach:
the linter is not very happy tho

.travis.yml
```diff
@@ -2,6 +2,7 @@ language: node_js
 node_js:
   - "8"
   - "10"
+  - "11"
```
imurchie: This is not really necessary. 11 is a stop-gap that will be EOL-ed in a couple of months.

In a couple of weeks, when 12 is released, we will update everything and drop support for 8.

PR #316, opened 2019-04-08
Fix the error which is thrown on download failure

mykola-mokhnach:
The `reject` argument is expected to be a valid error object

### DIFF
--- a/lib/basedriver/helpers.js
+++ b/lib/basedriver/helpers.js
@@ -204,41 +204,35 @@ async function configureApp (app, supportedAppExtensions) {
 }
 
 async function downloadApp (app, targetPath) {
-  let appUrl;
+  const {href} = url.parse(app);
+  const started = process.hrtime();
   try {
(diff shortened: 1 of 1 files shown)


lib/basedriver/helpers.js
```diff
@@ -204,41 +204,35 @@ async function configureApp (app, supportedAppExtensions) {
 }
 
 async function downloadApp (app, targetPath) {
-  let appUrl;
+  const {href} = url.parse(app);
```
dpgraham: This throws an exception if `app` isn't a `string` type. I think we should do some defensive validation here.

mykola-mokhnach: I think is fine, since downloadApp is only called from https://github.com/appium/appium-base-driver/pull/316/files#diff-42b9efd7ac57979fa56fdc25706f4ae5R77, where we are already doing the necessary validations

PR #320, opened 2019-04-24
Remove chromedriver dev dep

imurchie:
Weird thing... 

We have a fully functional Selenium server here already. No need to use Chromedriver or actually drive a browser!

Also, fo fun, removed unneeded dependencies.


mykola-mokhnach:
🎉

PR #325, opened 2019-05-12
added keepalive option to JWProxy

Dan-Maor:
Added an optional parameter to the JWProxy constructor to allow keeping the connection open as long as it is being used.

For example, enabling it for the connection to iOS' WebDriverAgent increases performance.

### DIFF
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -24,6 +24,7 @@ class JWProxy {
       base: '/wd/hub',
       sessionId: null,
       timeout: DEFAULT_REQUEST_TIMEOUT,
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
How about moving this parameter to server settings (aka command line arg). Or it makes sense to make it configurable on per session basis?

Dan-Maor:
Thing is - this way it will affect the value of every instance of JWProxy, including Android (which we saw previously was a problem in https://github.com/appium/appium/issues/11380).

I think that it would be safer to keep the value as false by default to avoid any unforeseen regressions and allow to override it in specific places where it is needed. Perhaps we can add the ability to explicitly disable it everywhere with a command line argument.

What do you think?

PR #328, opened 2019-06-04
Add endpoint that receives a webdriver script for batch command execution

jlipps:
The [batch command proposal](https://github.com/appium/appium/issues/11021) has been under discussion for some time. To summarize, the original proposal was for a new API that would essentially allow encoding of any Appium command in JSON format, including the use of magic values like `{{<element}}` to refer to an element found in a previous command. This original proposal had several drawbacks:

1. Quite a lot of code would need to be written on both client and server to support the feature. Most of the code would just be encoding/decoding commands from a JSON format.
2. This method interacts poorly with proxying, which is quite common in Appium tests. Say that a batched command set came in involving a context switch, such that subsequent commands in the batch are supposed to target chromedriver. The problem is that the batch command itself would be written on the driver, which can only awkwardly proxy it on.
3. Use of backreferences is just one example of all kinds of logic that users might wish to encode in batch commands, and be unable to. How would we handle conditionals, loops, and the like? We'd need to draw a line somewhere or invent a Turing complete mini language to support all the use cases!

This PR takes a much more radical approach. Basically, it asks that the user send in not a formatted list of commands, but rather a script (written in a supported library, in this case webdriverio). This script is run in a NodeJS VM with access to a driver object representing the current session. This proposal has a number of advantages:

1. Access to a full webdriver client
2. Access to a full programming language (JS) for use of logic
3. Very little code needing to be written on the client side (the Appium Java client would just assume someone will write a webdriverio script as a string. Or, the Java client could go crazy and implement a code transformation feature that outputs valid webdriverio code. Either way).
4. Very little code needing to be written on the server side. I think this PR is all that's necessary!

The main disadvantages of this new proposal are:

1. Theoretical security concerns running untrusted code. However, the code is sandboxed in a VM, and has no access to e.g. the `require` function. I'm having trouble thinking how a malicious user could abuse this feature, but it's possible I suppose.
2. We bring in a webdriver client (webdriverio) as a production dep. Thankfully it's not too big. In the future, this feature and its deps could be developed as a plugin/extension rather than part of Appium core.

I think this is pretty sweet, what do you all think?

### DIFF
--- a/lib/basedriver/commands/execute-child.js
+++ b/lib/basedriver/commands/execute-child.js
@@ -1,0 +1,128 @@
+import _ from 'lodash';
+import B from 'bluebird';
+import vm from 'vm';
+import log from '../logger';
+import { attach } from 'webdriverio';
+
+// duplicate defining these keys here so we don't need to re-load a huge appium
+// dependency tree into memory just to run a wdio script
+const W3C_ELEMENT_KEY = 'element-6066-11e4-a52e-4f735466cecf';
+const MJSONWP_ELEMENT_KEY = 'ELEMENT';
+
+async function runScript (driverOpts, script, timeout) {
+  // set up fake logger
+  const logLevels = ['error', 'warn', 'log'];
+  const logs = {};
+  const consoleFns = {};
+  for (const level of logLevels) {
+    logs[level] = [];
+    consoleFns[level] = (...logMsgs) => logs[level].push(...logMsgs);
+  }
+
+  const driver = attach(driverOpts);
+
+  const fullScript = buildScript(script);
+  // the timeout here will not matter really, but set it anyway to be on the
+  // safe side
+  const vmCtx = vm.runInNewContext(fullScript, {}, {timeout});
+
+  // run the driver script, giving user access to the driver object, a fake
+  // console logger, and a promise library
+  log.info('Running driver script in Node vm');
+  let result = await vmCtx(driver, consoleFns, B);
+
+  log.info('Ensuring driver script result is appropriate type for return');
+  result = coerceScriptResult(result);
+  return {result, logs};
+}
+
+/**
+ * Embed a user-generated script inside a method which takes only the
+ * predetermined objects we specify
+ *
+ * @param {string} script - the javascript to execute
+ *
+ * @return {string} - the full script to execute
+ */
+function buildScript (script) {
+  return `(async function execute (driver, console, Promise) {
+    ${script}
+  })`;
+}
+
+/**
+ * We can get any manner of crazy thing back from a vm executing untrusted
+ * code. We might also get WebdriverIO objects that aren't suitable for JSON
+ * response. So make sure we convert the things we know about to their
+ * appropriate response format, and squash other weird things.
+ *
+ * @param {Object} obj - object to convert and sanitize
+ *
+ * @return {Object} - safely converted object
+ */
+function coerceScriptResult (obj) {
+  // first ensure obj is of a type that can be JSON encoded safely. This will
+  // get rid of custom objects, functions, etc... and turn them into POJOs
+  try {
+    obj = JSON.parse(JSON.stringify(obj));
+  } catch (e) {
+    log.warn('Could not convert executeDriverScript to safe response!' +
+             `Result was: ${obj}. Will make it null`);
+    return null;
+  }
+
+  let res;
+
+  // now we begin our recursive case options
+  if (_.isPlainObject(obj)) {
+    // if we have an object, it's either an element object or something else
+    // webdriverio has no monadic object types other than element and driver,
+    // and we don't want to allow special casing return of driver
+    res = {};
+
+    if (obj[MJSONWP_ELEMENT_KEY] || obj[W3C_ELEMENT_KEY]) {
+      // if it's an element object, clear out anything that's not the key, and
+      // then return the object
+      if (obj[MJSONWP_ELEMENT_KEY]) {
+        res[MJSONWP_ELEMENT_KEY] = obj[MJSONWP_ELEMENT_KEY];
+      }
+
+      if (obj[W3C_ELEMENT_KEY]) {
+        res[W3C_ELEMENT_KEY] = obj[W3C_ELEMENT_KEY];
+      }
+      return res;
+    }
+
+    // otherwise, recurse into the object
+    for (const key of Object.keys(obj)) {
+      res[key] = coerceScriptResult(obj[key]);
+    }
+    return res;
+  }
+
+  // in the cae of an array, just recurse into the items
+  if (_.isArray(obj)) {
+    return obj.map(i => coerceScriptResult(i));
+  }
+
+  // base case, if it's not an object or array, return straightaway
+  return obj;
+}
+
+async function main (driverOpts, script, timeout) {
+  let res;
+  try {
+    res = {success: await runScript(driverOpts, script, timeout)};
+  } catch (error) {
+    res = {error: {message: error.message, stack: error.stack}};
+  }
+  await B.promisify(process.send, {context: process})(res);
+}
+
+if (require.main === module) {
+  log.info('Running driver execution in child process');
+  process.on('message', ({driverOpts, script, timeout}) => {
+    log.info('Parameters received from parent process');
+    main(driverOpts, script, timeout);
+  });
+}
--- a/lib/basedriver/commands/execute-child.js
+++ b/lib/basedriver/commands/execute-child.js
@@ -10,6 +10,10 @@ const W3C_ELEMENT_KEY = 'element-6066-11e4-a52e-4f735466cecf';
 const MJSONWP_ELEMENT_KEY = 'ELEMENT';
 
 async function runScript (driverOpts, script, timeout) {
+  if (!_.isNumber(timeout)) {
+    throw new Error('Timeout parameter must be a number');
+  }
+
   // set up fake logger
   const logLevels = ['error', 'warn', 'log'];
   const logs = {};
--- a/lib/basedriver/commands/execute.js
+++ b/lib/basedriver/commands/execute.js
@@ -5,7 +5,7 @@ import { attach } from 'webdriverio';
 import { MJSONWP_ELEMENT_KEY, W3C_ELEMENT_KEY } from '../../protocol/protocol';
 
 const BASE_PATH = '/wd/hub'; // TODO defining this here is brittle but it is hardcoded in routes, so...
-const FEAT_FLAG = 'execute-driver-script';
+const FEAT_FLAG = 'execute_driver_script';
 const SCRIPT_TYPE_WDIO = 'webdriverio';
 // TODO add wd script type
 
--- a/lib/basedriver/commands/execute.js
+++ b/lib/basedriver/commands/execute.js
@@ -34,6 +34,14 @@ commands.executeDriverScript = async function (script, scriptType = 'webdriverio
                     'is required. This is probably a programming error in the driver');
   }
 
+  const logLevels = ['error', 'warn', 'log'];
+  const logs = {};
+  const consoleFns = {};
+  for (const level of logLevels) {
+    logs[level] = [];
+    consoleFns[level] = (...logMsgs) => logs[level].push(...logMsgs);
+  }
+
   const driver = attach({
     sessionId: this.sessionId,
     protocol: 'http', // Appium won't ever be behind ssl locally
@@ -44,12 +52,13 @@ commands.executeDriverScript = async function (script, scriptType = 'webdriverio
     capabilities: this.caps
   });
   const fullScript = buildScript(script);
-  const result = await vm.runInNewContext(fullScript)(driver);
-  return coerceScriptResult(result);
+  let result = await vm.runInNewContext(fullScript)(driver, consoleFns);
+  result = coerceScriptResult(result);
+  return {result, logs};
 };
 
 function buildScript (script) {
-  return `(async function execute (driver) {
+  return `(async function execute (driver, console) {
     ${script}
   })`;
 }
--- a/lib/basedriver/commands/execute.js
+++ b/lib/basedriver/commands/execute.js
@@ -1,13 +1,14 @@
 import _ from 'lodash';
+import path from 'path';
+import cp from 'child_process';
 import log from '../logger';
-import vm from 'vm';
-import { attach } from 'webdriverio';
-import { MJSONWP_ELEMENT_KEY, W3C_ELEMENT_KEY } from '../../protocol/protocol';
+import B from 'bluebird';
 
 const BASE_PATH = '/wd/hub'; // TODO defining this here is brittle but it is hardcoded in routes, so...
 const FEAT_FLAG = 'execute_driver_script';
+const DEFAULT_SCRIPT_TIMEOUT = 1000 * 60 * 60; // default to 1 hour timeout
 const SCRIPT_TYPE_WDIO = 'webdriverio';
-// TODO add wd script type
+// TODO add wd script type at some point
 
 let commands = {};
 
@@ -23,7 +24,9 @@ let commands = {};
  * @returns {Object} - a JSONifiable object representing the return value of
  * the script
  */
-commands.executeDriverScript = async function (script, scriptType = 'webdriverio') {
+commands.executeDriverScript = async function (script, scriptType = 'webdriverio',
+  timeout = DEFAULT_SCRIPT_TIMEOUT) {
+
   if (!this.isFeatureEnabled(FEAT_FLAG)) {
     throw new Error(`Execute driver script functionality is not available ` +
                     `unless server is started with --allow-insecure including ` +
@@ -39,15 +42,11 @@ commands.executeDriverScript = async function (script, scriptType = 'webdriverio
                     'is required. This is probably a programming error in the driver');
   }
 
-  const logLevels = ['error', 'warn', 'log'];
-  const logs = {};
-  const consoleFns = {};
-  for (const level of logLevels) {
-    logs[level] = [];
-    consoleFns[level] = (...logMsgs) => logs[level].push(...logMsgs);
+  if (!_.isNumber(timeout)) {
+    throw new Error('Timeout parameter must be a number');
   }
 
-  const driver = attach({
+  const driverOpts = {
     sessionId: this.sessionId,
     protocol: 'http', // Appium won't ever be behind ssl locally
     hostname: this.opts.host,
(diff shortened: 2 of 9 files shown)


lib/basedriver/commands/execute.js
```diff
@@ -0,0 +1,102 @@
...
+    sessionId: this.sessionId,
+    protocol: 'http', // Appium won't ever be behind ssl locally
+    hostname: this.opts.host,
+    port: this.opts.port,
+    path: BASE_PATH,
+    isW3C: true, // TODO should probably retrieve client protocol and declare the same one for the script
+    capabilities: this.caps
+  });
+  const fullScript = buildScript(script);
+  const result = await vm.runInNewContext(fullScript)(driver);
```
mykola-mokhnach: Do we have access to the server's file system from this script? If yes then I wouldn't appreciate this approach much

lib/basedriver/commands/execute.js
```diff
@@ -0,0 +1,102 @@
...
+  return coerceScriptResult(result);
+};
+
+function buildScript (script) {
+  return `(async function execute (driver) {
```
mykola-mokhnach: why it is necessary to wrap the script into a function?

lib/basedriver/commands/execute.js
```diff
@@ -0,0 +1,102 @@
...
+ * a new nodejs VM, and which has available a webdriverio driver object, having
+ * already been attached to the currently running session.
+ */
+commands.executeDriverScript = async function (script, scriptType) {
+  if (!!scriptType && scriptType !== 'webdriverio') {
+    throw new Error("Only the 'webdriverio' script type is currently supported");
+    // TODO should probably add support for WD script type as well?
+  }
+
+  if (!this.opts.host || !this.opts.port) {
```
mykola-mokhnach: isn't the host always equal to localhost?

lib/basedriver/commands/execute.js
```diff
@@ -0,0 +1,102 @@
...
+    throw new Error('Host or port of running server were not defined; this is required');
+  }
+  const driver = attach({
+    sessionId: this.sessionId,
+    protocol: 'http', // Appium won't ever be behind ssl locally
+    hostname: this.opts.host,
+    port: this.opts.port,
+    path: BASE_PATH,
+    isW3C: true, // TODO should probably retrieve client protocol and declare the same one for the script
```
mykola-mokhnach: 👍 for TODO

lib/basedriver/commands/execute.js
mykola-mokhnach: how are we going to show errors caused by script interpretation?

mykola-mokhnach: how can we guarantee that this piece of code won't loop forever like it is described in https://nodejs.org/api/vm.html#vm_timeout_limitations_when_using_process_nexttick_promises_and_queuemicrotask?

lib/basedriver/commands/execute.js
```diff
@@ -0,0 +1,102 @@
...
+ * We can get any manner of crazy thing back from a vm executing untrusted
+ * code. We might also get WebdriverIO objects that aren't suitable for JSON
+ * response. So make sure we convert the things we know about to their
+ * appropriate response format, and squash other weird things.
+ */
+function coerceScriptResult (obj) {
+  // first ensure obj is of a type that can be JSON encoded safely. This will
+  // get rid of custom objects, functions, etc... and turn them into POJOs
+  try {
+    obj = JSON.parse(JSON.stringify(obj));
```
mykola-mokhnach: this also won't work for bigint number

test/basedriver/driver-e2e-tests.js
```diff
@@ -245,6 +262,110 @@ function baseDriverE2ETests (DriverClass, defaultCaps = {}) {
...
+          const timeouts = await driver.getTimeouts();
+          const status = await driver.status();
+          return [timeouts, status];
+        `;
+        const res = await request({
+          url: `http://localhost:8181/wd/hub/session/${sessionId}/appium/execute_driver`,
+          method: 'POST',
+          json: {script, type: 'webdriverio'},
+        });
+        const expectedTimeouts = {'command': 250, 'implicit': 0};
```
mykola-mokhnach: should key names be quoted?

lib/basedriver/commands/execute.js
```diff
@@ -0,0 +1,102 @@
...
+let commands = {};
+
+/*
+ * TODO fill out entire docstring
+ *
+ * This method takes a string which is executed as javascript in the context of
+ * a new nodejs VM, and which has available a webdriverio driver object, having
+ * already been attached to the currently running session.
+ */
+commands.executeDriverScript = async function (script, scriptType) {
```
mykola-mokhnach: should scriptType have a default value here?

lib/basedriver/commands/execute.js
jlipps: We can add a timeout as a parameter for the vm. And code run in the vm doesn't have access to the filesystem since it has no access to `require`, so you can't e.g. `require('fs')`.

Errors caused by script interpretation are thrown and sent to the client. I should probably add a test for errors that cause the script not to even compile, but I bet it works with the code as is.

lib/basedriver/commands/execute.js
jlipps: I want to create the driver outside the VM, and then pass the driver into the VM, so that the VM itself has completely minimal access to everything.

lib/basedriver/commands/execute.js
jlipps: it doesn't have to be, you can set it with the `-h` parameter.

test/basedriver/driver-e2e-tests.js
jlipps: nope, guess i was thinking in python

lib/basedriver/commands/execute.js
jlipps: i need to read the router code but i think it will pass in a value here regardless of what we do?

lib/basedriver/commands/execute.js
jlipps: i'm ok documenting that any return value must be consist of simple valid JSON values or elements. if someone is working with bigints they can also do the conversion in the script and return a regular int or some other object representation. i cared about supporting elements as a really common use case.

xcolwell:
This is very nice direction for remote device testing and distributed device testing. Having a way to optimize performance critical sections like "find and act" while running a full script on the client has not been possible with Appium.

A couple questions:

- how are remote exceptions returned to the client?
- how are stdout/stderr returned to the client?
- can the script return a json to the client?
- is there a watchdog on the uploaded script for example max cpu time per script, max elapsed time per script?
- do you see a standard set of available modules in the node vm? In other words is there a standard "header" file that gets appended to each batch script?
- probably this would be enabled with a special flag? I think the semantics of `--relaxed-security` make sense here, but it would be most useful for us to enable just this batch feature without the other features guarded by this flag. 

Also I think it would be useful to pass the script language with the script, so that cloud providers can choose to optionally support transpilation of other languages to javascript.

KazuCocoa:
a note after reading https://github.com/appium/appium-base-driver/pull/328/files#diff-aa816d90f6389bb3ff7a3fc331643232R287 ...

an idea of how to pass valid JS commands to the server. (To reduce mistake/work load in client side)

like Selenium, we'll prepare some JS commands as atom commands. Each client gets, reads and sends it to server like https://github.com/SeleniumHQ/selenium/blob/2abd80f236d1a7459ef638e96af8c4efd86b4abd/rb/lib/selenium/webdriver/atoms.rb#L30 (we can fix the JS file to fix some issues instead o updating all clients. This is a behefit of shared code.)

If sending command is complicated, this method does not work. Users should write JS (or will we implement each languages->JS converter?).

jlipps:
@KazuCocoa I'm not totally sure I'm following you. Are you saying that it's a bad idea to give clients the ability to write full-on javascript? And instead we implement client methods that simply generate correct javascript?

I think that's a good idea, but I don't see them as mutually exclusive. You could either write something like this:

```js
await driver.executeDriverScript(`
    const el = await driver.$("~foo");
    await el.click();
`);
```

or, you could write this (if the client supports it, let's say we're using python:

```py
with driver.batch() as batch:
    el = batch.find_element_by_accessibility_id("foo")
    el.click()
```

And then the python client would, as part of interpreting the context manager there, turn that into valid webdriverio code underneath the hood, and call `driver.execute_script` on its own?

Is that what you're proposing? If so I agree that'd be awesome, and then it's just up to the clients to implement this "safe" way of using executeScript. In practice though, I'm not sure how easy it would be to do this translation. What about this script?

```py
with driver.batch() as batch:
    els = batch.find_elements_by_accessibility_id("foo")
    for el in els:
        el.click()
```

How do you translate the `for` loop into JS? You'd need to do the translation based on a Python AST or something, which is getting kinda crazy at this point.

jlipps:
thanks for the questions, @xcolwell!


> how are remote exceptions returned to the client?

Same as with any other error

> how are stdout/stderr returned to the client?

Right now they are not (and running inside a pure JS VM, there's no such thing as stdout/stderr per se), but we could make some log object available to the script, which could then be used to put the output in the appium log as well as return it as part of the return value of the function.

> can the script return a json to the client?

Yes, the script can return any simple javascript object at any level of complexity, and JSON is the transport method.

> is there a watchdog on the uploaded script for example max cpu time per script, max elapsed time per script?

No, I'm not sure if it's possible to attach a watchdog to a VM, but this could presumably be done of Appium as a whole. I'm assuming actions in the VM get counted on a system resource level are counted as happening within the appium process

> do you see a standard set of available modules in the node vm? In other words is there a standard "header" file that gets appended to each batch script?

There could be! We'd just need to decide what we want to make available to the user. In this PR it's just the driver object. But we mentioned above maybe a log object also makes sense. What else makes sense? Within the constraints of security.

> probably this would be enabled with a special flag? I think the semantics of --relaxed-security make sense here, but it would be most useful for us to enable just this batch feature without the other features guarded by this flag.

I agree.

jlipps:
@imurchie any objections to this approach?

lib/basedriver/commands/execute.js
christian-bromann: @jlipps you can find all supported environment flags here: https://github.com/webdriverio/webdriverio/blob/master/packages/webdriver/src/utils.js#L301-L310

lib/basedriver/driver.js
```diff
@@ -243,6 +243,35 @@ class BaseDriver extends Protocol {
...
+    // if we specifically have allowed the feature, return true
+    if (this.allowInsecure && this.allowInsecure.includes(name)) {
+      return true;
+    }
+
+    // if we've globally allowed insecure features
+    if (this.relaxedSecurityEnabled) {
+
+      // but have explicitly denied this one, return false
+      if (this.denyInsecure && this.denyInsecure.includes(name)) {
```
mykola-mokhnach: denyInsecure should have the highest priority

lib/basedriver/driver.js
```diff
@@ -243,6 +243,35 @@ class BaseDriver extends Protocol {
...
+  /**
+   * Check whether a given feature is enabled via its name
+   *
+   * @param {string} name - name of feature/command
+   *
+   * @returns {Boolean}
+   */
+  isFeatureEnabled (name) {
+    // if we specifically have allowed the feature, return true
+    if (this.allowInsecure && this.allowInsecure.includes(name)) {
```
mykola-mokhnach: you can use _.include, which is more flexible while handling null/undefined args

jlipps:
ok @mykola-mokhnach `isFeatureEnabled` has been updated, and unit tests added. i think this code is now all good to go. i'll be happy to find uses of `relaxedSecurityEnabled` in the codebase and update to use this new method.

jlipps:
Alright, I've made all the downstream changes so once we merge all of these, the whole codebase should be using the new security model.

lib/basedriver/commands/execute.js
```diff
@@ -0,0 +1,139 @@
...
+    protocol: 'http', // Appium won't ever be behind ssl locally
+    hostname: this.opts.host,
+    port: this.opts.port,
+    path: BASE_PATH,
+    isW3C: this.isW3CProtocol(),
+    isMobile: true,
+    capabilities: this.caps
+  });
+  const fullScript = buildScript(script);
+  let result = await vm.runInNewContext(fullScript)(driver, consoleFns);
```
mykola-mokhnach: so, do we want to limit the maximum script execution time?

jlipps: I think it's a good idea, though an infinite loop will not be prevented by doing so. I guess this should be modulated by a cap?

mykola-mokhnach: This can be also an argument to the endpoint. But it also makes sense to have a default value of, let say 60 minutes.

jlipps: oh yeah, endpoint argument makes more sense.

jlipps: ok i spent a bunch of time trying to implement this, and it doesn't work, because the client script is executed in a promise, so it avoids the node vm timeout. i tried to implement timeouts using promises, but Promise.race causes the whole process to hang presumably because one of the promises is never read. i'm not sure. I think we have 2 options:

1) forget any useful timeout argument and just make clear to the sys admin that a single user could cause an appium server to stall using this

2) run all this in a subproc, not a nodejs vm, so we can kill it from the outside if it's being bad.

what do you think?

jlipps: i decided to work on a combination of proc fork + vm, we'll see...

jlipps: ok, i got it working. this is a bit slower since it has to spin up a new node process for each request, but it will be vastly safer i guess!

jlipps: (it will even work if someone implements an infinite loop)

jlipps:
will work on docs in another PR in the main repo

jlipps:
OK, docs added in https://github.com/appium/appium/pull/12807

lib/basedriver/commands/execute-child.js
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * Embed a user-generated script inside a method which takes only the
+ * predetermined objects we specify
+ *
+ * @param {string} script - the javascript to execute
+ *
+ * @return {string} - the full script to execute
+ */
+function buildScript (script) {
+  return `(async function execute (driver, console, Promise) {
```
mykola-mokhnach: aren't console and Promise shadowing standard object names?

lib/basedriver/commands/execute-child.js
```diff
@@ -0,0 +1,128 @@
...
+import vm from 'vm';
+import log from '../logger';
+import { attach } from 'webdriverio';
+
+// duplicate defining these keys here so we don't need to re-load a huge appium
+// dependency tree into memory just to run a wdio script
+const W3C_ELEMENT_KEY = 'element-6066-11e4-a52e-4f735466cecf';
+const MJSONWP_ELEMENT_KEY = 'ELEMENT';
+
+async function runScript (driverOpts, script, timeout) {
```
mykola-mokhnach: should timeout have a default value?

lib/basedriver/commands/execute-child.js
jlipps: these standard object names are not available in the VM. it's literally just a JS language interpreter, it has no standard web or node objects available, so we have to provide everything.

lib/basedriver/commands/execute-child.js
jlipps: it does, in the important place where the request from the user is handled (in `execute-child.js`) but i'll add a check to make sure it's included

lib/basedriver/commands/execute.js
jlipps: Hmm, I'm not sure what's going on here. If you increase the timeout does it just wait until the timeout? 1s is a pretty low timeout. The issue appears to be that the XCUITest driver is proxying the /timeouts call to WDA, but WDA can't handle it. Probably /timeouts needs to be added to NO_PROXY in WDA?

But the more concerning issue is why we got a timeout error and not the "Unhandled Endpoint" error which we would have expected. That's why I'm asking you to increase the timeout in case it's just a race condition.

jlipps: > Will you work to set this.opts.host and this.opts.port in appium repo?

I don't understand. My assumption was that `this.opts.host` and `this.opts.port` are always set because the main Appium server copies its args on to `InnerDriver.opts`, is that not true?

KazuCocoa: > this.opts.host and this.opts.port are always set

mm, let me take a look further later.
On my environment, 

1. `run clean` in appium repo
2. Link the node_module/appium-base-driver to this PR
3. Launch the appium
4. Run client script as xcuitest

Then, _Host or port of running server were not defined_ error was raised. I set `this.opts.host = 'localhost'; this.opts.port = '4723';` before this line, it could avoid. So, I thought this option was new items.
I did not check other driver cases.

jlipps: were you starting an Appium server or just a XCUITest driver server? The latter will not work; only the main Appium server copies the CLI args to the inner drivers.

KazuCocoa: I ran appium server with `node . --relaxed-security` in appium repo (and xcuitest was automationName)
ah, did I miss some server arguments?

jlipps:
Also just realized that to support this feature, `/execute_driver` has to be added to NO_PROXY for specific drivers.

lib/basedriver/commands/execute.js
KazuCocoa: ah, what about `opts.address` instead of `opts.host`?
Let me check...
https://github.com/appium/appium/search?q=address&unscoped_q=address

KazuCocoa: When I replaced `opts.host` with `opts.address` in _lib/basedriver/commands/execute.js_, it worked fine with Python client script.
I could configure them as `-p 4723 -a localhost` [server arg](https://github.com/appium/appium/blob/master/docs/en/writing-running-appium/server-args.md).

Could you double check about _address_ or _host_?

jlipps: OK, i ran into this too. I guess it's a bad assumption.

jlipps:
I've run some functional tests with @KazuCocoa 's python client changes, and they work (though I ran into a bug with webdriverio sendkeys, seems like [these lines](https://github.com/webdriverio/webdriverio/blob/6d2894d8be817816ca6c841c2a1ea19fe3f3fec3/packages/webdriverio/src/commands/element/addValue.js#L31-L33) are causing the problem--cc @christian-bromann )

christian-bromann:
> though I ran into a bug with webdriverio sendkeys

I will look into it.

jlipps:
will wait a bit to see what @christian-bromann says before merging, in case a code change is required on this end. otherwise i think this is good to go, and will merge/publish all the dependent PRs once this one is merged.

jlipps:
ok, with the new version of webdriverio, i've confirmed this works, so will merge now!

jlipps:
published in 3.17.0

PR #333, opened 2019-06-20
Add parseCapsArray to base driver

vrunoa:
Adding parseArray to appium-base-driver so it can be used on appium-android-driver and appium-xcuitest-driver.


lib/basedriver/helpers.js
```diff
@@ -357,6 +357,27 @@ function duplicateKeys (input, firstKey, secondKey) {
...
 }
 
+/**
+ * Takes a desired capability and tries to JSON.parse it as an array,
+ * and either returns the parsed array or a singleton array.
+ *
+ * @param {any} cap A desired capability
+ */
+function parseArray (cap) {
+  let parsedCaps;
```
mykola-mokhnach: I'd rather select some more generic var name

lib/basedriver/helpers.js
```diff
@@ -357,6 +357,27 @@ function duplicateKeys (input, firstKey, secondKey) {
   return input;
 }
 
+/**
+ * Takes a desired capability and tries to JSON.parse it as an array,
+ * and either returns the parsed array or a singleton array.
+ *
+ * @param {any} cap A desired capability
```
mykola-mokhnach: missing throws and return descriptions

`@param` is expected to be `{string|Array<String>}`

mykola-mokhnach:
How about haiving this method in appium-support/utils instead?

vrunoa:
@mykola-mokhnach it's fine by me, @imurchie proposed we move this method here since I want to use it in `appium-xcuitest-driver` and `appium-android-driver`. Let me know what you guys think it's the best place.

imurchie:
I suggested here because it _is_ capability specific. Perhaps a name change (`parseCapArray` or something) would indicate that.

lib/basedriver/helpers.js
mykola-mokhnach: or, like @imurchie mentioned, rename the function to reflect the fact it is used for capabilities

lib/basedriver/helpers.js
```diff
@@ -357,6 +357,27 @@ function duplicateKeys (input, firstKey, secondKey) {
...
+ * Takes a desired capability and tries to JSON.parse it as an array,
+ * and either returns the parsed array or a singleton array.
+ *
+ * @param {string|Array<String>} cap A desired capability
+ */
+function parseCapsArray (cap) {
+  let parsedCaps;
+  try {
+    parsedCaps = JSON.parse(cap);
+  } catch (ign) { }
```
umutuzgur: Could you add a comment here why we are ignoring the exception? I understood your reason after some time but it would be good to have it documented here

vrunoa: @umutuzgur take a look at this new approach for re review

PR #334, opened 2019-06-28
Optimize memory usage by proxy class

mykola-mokhnach:
Sometimes proxied requests may return really big objects. Calling JSON.parse/stringify on them multiple times is really expensive from both memory and CPU usage perspective.

I've also removed the redirect case for session creation (return code 303), since I didn't find any mentions of this case in specs. Please let me know if this is really needed, so I can put it back.

### DIFF
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -132,8 +132,14 @@ class JWProxy {
       forever: this.keepAlive,
     };
     if (util.hasValue(body)) {
-      if (!_.isPlainObject(body)) {
-        body = JSON.parse(body);
+      if (typeof body !== 'object') {
+        try {
+          body = JSON.parse(body);
+        } catch (e) {
+          throw new Error('Cannot interpret the request body as valid JSON: ' +
+            _.truncate(_.isString(body) ? body : JSON.stringify(body),
+            {length: LOG_OBJ_LENGTH}));
+        }
       }
       reqOpts.json = body;
     }
@@ -144,8 +150,8 @@ class JWProxy {
     }
 
     log.debug(`Proxying [${method} ${url || '/'}] to [${method} ${newUrl}] ` +
-             (body ? `with body: ${_.truncate(_.isString(body) ? body : JSON.stringify(body),
-              {length: LOG_OBJ_LENGTH})}` : 'with no body'));
+      (body ? `with body: ${_.truncate(_.isString(body) ? body : JSON.stringify(body),
+      {length: LOG_OBJ_LENGTH})}` : 'with no body'));
 
     const throwProxyError = (error) => {
       const message = `The request to ${url} has failed`;
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -134,14 +134,15 @@ class JWProxy {
     if (util.hasValue(body)) {
       if (typeof body !== 'object') {
         try {
-          body = JSON.parse(body);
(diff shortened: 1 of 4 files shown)


lib/jsonwp-proxy/proxy.js
```diff
@@ -144,46 +150,55 @@ class JWProxy {
...
+      throw err;
+    };
     try {
-      res = await this.request(reqOpts);
-      resBody = res.body;
-      log.debug(`Got response with status ${res.statusCode}: ${_.truncate(JSON.stringify(resBody), {length: LOG_OBJ_LENGTH})}`);
-      if (/\/session$/.test(url) && method === 'POST') {
-        if (res.statusCode === 200) {
-          this.sessionId = resBody.sessionId;
-        } else if (res.statusCode === 303) {
```
jlipps: you are intentionally removing support for the older redirect style of session start?

mykola-mokhnach: I actually wanted to know if it is still used anywhere. As far as I can see this is now obsolete

jlipps: are we sure no supported chromedrivers do this? i remember we had to add it for one of the drivers we proxy to, i'm not sure which one.

mykola-mokhnach: I would say it would be smarter to make `request` itself to follow redirects rather than adding a special if case. Redirects should be transparent for the proxy client

jlipps: makes sense to me.

lib/jsonwp-proxy/proxy.js
```diff
@@ -131,9 +131,15 @@ class JWProxy {
...
-    if (body !== null) {
+    if (util.hasValue(body)) {
       if (typeof body !== 'object') {
-        body = JSON.parse(body);
+        try {
+          body = JSON.parse(body);
+        } catch (e) {
+          throw new Error('Cannot interpret the request body as valid JSON: ' +
+            _.truncate(_.isString(body) ? body : JSON.stringify(body),
+            {length: LOG_OBJ_LENGTH}));
```
rkavalap: Could you please not truncate this if it was in debug or verbose mode.

mykola-mokhnach: which driver property points to the fact that debug mode is enabled? I don't know such

lib/jsonwp-proxy/proxy.js
imurchie: > I would say it would be smarter to make request itself to follow redirects rather than adding a special if case. Redirects should be transparent for the proxy client

Has this been done, then?

mykola-mokhnach: no, I didn't spend time on this. I assume this code is obsolete now and is not used anywhere. In general request promise should throw an error if POST request request return 300+ code: https://github.com/request/request-promise/issues/34

I'm going to take care about possible consequences in beta if there are any, but for now I don't expect such

imurchie: Ok. I guess if this gets in and we watch for odd behaviours, I'm ok.

lib/jsonwp-proxy/proxy.js
rkavalap: https://github.com/appium/appium-base-driver/pull/72/files#issuecomment-245351191

mykola-mokhnach: That PR keep truncate and I know why. Removing it completely will quickly overload the log buffer with megabytes-long strings and cause an out of memory error

PR #336, opened 2019-07-18
Update the logic that determines the session protocol

mykola-mokhnach:
It might be that we proxy /status call first to the downstream driver and it responds in W3C style (since this is a default setting), but then after createSession request is sent the photo is changed to JSONWP based on the provided caps. This change should properly handle such scenario.


mykola-mokhnach:
Makes sense. Added the explanation comment

PR #337, opened 2019-07-19
Improve the logic of proxyReqRes method

mykola-mokhnach:
`proxyCommand` call already returns an object, so there is no need to perform double conversion there. Also, it only makes sense to call stringily if the body object has been modified to save resources.

### DIFF
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -166,8 +166,8 @@ class JWProxy {
       const res = await this.request(reqOpts);
       // `res.body` might be really big
       // Be careful while handling it to avoid memory leaks
-      const resObj = util.safeJsonParse(res.body);
-      if (!_.isPlainObject(resObj)) {
+      const resBodyObj = util.safeJsonParse(res.body);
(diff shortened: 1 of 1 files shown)


lib/jsonwp-proxy/proxy.js
```diff
@@ -299,25 +299,28 @@ class JWProxy {
...
 
     res.headers = response.headers;
     res.set('content-type', response.headers['content-type']);
     // if the proxied response contains a sessionId that the downstream
     // driver has generated, we don't want to return that to the client.
     // Instead, return the id from the request or from current session
-    body = util.safeJsonParse(body);
-    if (body && body.sessionId) {
+    let isResBodyModified = false;
+    if (_.has(resBodyObj, 'sessionId')) {
```
jlipps: could there ever be a situation where `resBodyObj` is a JSON string rather than an object here?

mykola-mokhnach: isPlainObject check should take care of this

mykola-mokhnach: as for now proxyCommand will always return an object as the second argument

PR #341, opened 2019-07-31
Handle Content-Type with parameter

sa-spag:
Hey there,

I figured out HockeyApp responds with `Content-Type: application/zip; charset=utf-8` (which I agree makes no sense). I suggest we relax the Content-Type header test to handle such cases.

### DIFF
--- a/lib/basedriver/helpers.js
+++ b/lib/basedriver/helpers.js
@@ -110,7 +110,7 @@ async function configureApp (app, supportedAppExtensions) {
       if (headers['content-type']) {
         logger.debug(`Content-Type: ${headers['content-type']}`);
         // the filetype may not be obvious for certain urls, so check the mime type too
-        if (ZIP_MIME_TYPES.includes(headers['content-type'])) {
+        if (ZIP_MIME_TYPES.some(mimeType => new RegExp(`\\b${_.escapeRegExp(mimeType)}\\b`).test(headers['content-type']))) {
           if (!fileName) {
             fileName = `${DEFAULT_BASENAME}.zip`;
           }
--- a/test/basedriver/helpers-e2e-specs.js
+++ b/test/basedriver/helpers-e2e-specs.js
@@ -1,5 +1,6 @@
 import chai from 'chai';
 import path from 'path';
+import url from 'url';
 import chaiAsPromised from 'chai-as-promised';
 import { fs } from 'appium-support';
 import { configureApp } from '../../lib/basedriver/helpers';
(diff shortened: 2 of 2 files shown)


lib/basedriver/helpers.js
```diff
@@ -110,7 +110,7 @@ async function configureApp (app, supportedAppExtensions) {
       if (headers['content-type']) {
         logger.debug(`Content-Type: ${headers['content-type']}`);
         // the filetype may not be obvious for certain urls, so check the mime type too
-        if (ZIP_MIME_TYPES.includes(headers['content-type'])) {
+        if (ZIP_MIME_TYPES.some(mimeType => headers['content-type'].includes(mimeType))) {
```
mykola-mokhnach: how about

```
new RegExp(`\\b${_.escapeRegExp(mimeType)}\\b`).test(headers['content-type'])
```

test/basedriver/helpers-e2e-specs.js
```diff
@@ -86,8 +86,10 @@ describe('app download and configuration', function () {
             // for testing zip file content types
             if (req.url.indexOf('mime-zip') !== -1) {
               res.setHeader('content-type', 'application/zip');
-            } else if (req.url.indexOf('mime-bip') !== 1) {
+            } else if (req.url.indexOf('mime-bip') !== -1) {
               res.setHeader('content-type', 'application/bip');
+            } else if (req.url.indexOf('mime-parameter') !== -1) {
```
mykola-mokhnach: I think `includes` would be more readable here in comparison to indexOf

mykola-mokhnach:
Thanks for the fix @sa-spag

test/basedriver/helpers-e2e-specs.js
sa-spag: I refactored the whole thing by passing the actual header value as a query parameter.

lib/basedriver/helpers.js
```diff
@@ -110,7 +110,7 @@ async function configureApp (app, supportedAppExtensions) {
       if (headers['content-type']) {
         logger.debug(`Content-Type: ${headers['content-type']}`);
         // the filetype may not be obvious for certain urls, so check the mime type too
-        if (ZIP_MIME_TYPES.includes(headers['content-type'])) {
+        if (ZIP_MIME_TYPES.some(mimeType => new RegExp(`\\b${_.escapeRegExp(mimeType)}\\b`).test(headers['content-type']))) {
```
mykola-mokhnach: we could even add an option to ignore case here to cover the craziest content types :)

test/basedriver/helpers-e2e-specs.js
```diff
@@ -84,10 +85,9 @@ describe('app download and configuration', function () {
               return;
             }
             // for testing zip file content types
-            if (req.url.indexOf('mime-zip') !== -1) {
-              res.setHeader('content-type', 'application/zip');
-            } else if (req.url.indexOf('mime-bip') !== 1) {
-              res.setHeader('content-type', 'application/bip');
+            const contentType = new URLSearchParams(url.parse(req.url).search).get('content-type');
```
mykola-mokhnach: is this doing the same as before?

sa-spag: Tests may now pass a `content-type` query parameter so the test server sets the response Content-Type header accordingly. I reflected this change in existing tests.

PR #343, opened 2019-08-07
Fix WinAppDriver's missing "value" property issue

ndimer:
Add 'value' property to the response body's JSON, if no 'value' property was received in the original response.

### DIFF
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -172,6 +172,11 @@ class JWProxy {
         // If it cannot be coerced to an object then the response is wrong
         throwProxyError(res.body);
       }
+      // add 'value' property if missing, to comply with W3C standard
+      // fixes missing 'value' property issue in WinAppDriver: https://github.com/webdriverio/webdriverio/issues/3807
+      if (!_.has(resObj, 'value')) {
+        resObj.value = null;
+      }
       log.debug(`Got response with status ${res.statusCode}: ` +
         _.truncate(_.isString(res.body) ? res.body : JSON.stringify(res.body), {
           length: LOG_OBJ_LENGTH,


lib/jsonwp-proxy/proxy.js
```diff
@@ -172,6 +172,11 @@ class JWProxy {
         // If it cannot be coerced to an object then the response is wrong
         throwProxyError(res.body);
       }
+      // add 'value' property if missing, to comply with W3C standard
+      // fixes missing 'value' property issue in WinAppDriver: https://github.com/webdriverio/webdriverio/issues/3807
+      if (!_.has(resObj, 'value')) {
```
dpgraham: This is throwing a lint error for some odd reason. It's saying `resObj` is not defined :/

dpgraham:
@ndimer you'll need to sign this before we can merge this: https://cla.js.foundation/appium/appium-base-driver?pullRequest=343

lib/jsonwp-proxy/proxy.js
dpgraham: @ndimer I think you may need to rebase again. It's `resBodyObj` now: https://github.com/appium/appium-base-driver/blob/master/lib/jsonwp-proxy/proxy.js#L169

mykola-mokhnach: I don't think this is the right place to put this. The response must be fixed after the actual protocol is determined.

mykola-mokhnach: most likely in the appium umbrella driver

dpgraham: @mykola-mokhnach Could we move it down to after the conditional on line 207 (it won't let me link to that line). 

```javascript
If (this..downstreamProtocol === 'W3C' && !_.has(resObj, 'value')) {
  resObj.value = null
}
```

dpgraham: Actually, looking at it again, both MJSONWP and W3C send `value` parameter. So I think it should be acceptable to just set it to `null` if value not found.

dpgraham: I think setting it to `null` should be okay.

This method will be unaffected by setting the value to null

```
getProtocolFromResBody (resObj) {
    if (util.hasValue(resObj.status)) {
      return MJSONWP;
    }
    if (util.hasValue(resObj.value)) {
      return W3C;
    }
  }
```

dpgraham: MJSONWP and W3C both have a 'value' key in the return object.

mykola-mokhnach: > This method will be unaffected by setting the value to null

Yes and now.

The purpose of this PR is to fix responses conversion into W3C format. Although, no conversion is happening here. We even don't know what the upstream protocol is. 

That is why I would rather put this fix into the place where the actual response scheme is converted before being sent to the API client (protocol.js -> buildHandler most likely).

ndimer: So what is the bottom line? Will one of you put it in the right place or shall we leave it where I originally added it?

mykola-mokhnach: I'll prepare the PR soon

PR #346, opened 2019-08-14
stop hardcoding /wd/hub and make route prefix configurable

jlipps:
We had /wd/hub all over the place, which was ugly and outdated. Turned out to be quite a lot of work to generalize an endpoint base path, but I think this will work. Help me make sure this won't break anything!


test/jsonwp-proxy/proxy-specs.js
```diff
@@ -1,186 +1,195 @@
...
-    it('should proxy strange responses', async function () {
-      let j = mockProxy({sessionId: '123'});
-      let [req, res] = buildReqRes('/nochrome', 'GET');
-      await j.proxyReqRes(req, res);
-      res.sentCode.should.equal(100);
-      res.sentBody.should.eql({status: 0, value: {message: 'chrome not reachable'}});
-    });
-  });
-});
+// transpile:mocha
```
jlipps: there were a bunch of \r characters at the end of lines in this file for some reason, and when i removed them the whole file got readded to git. not sure why.

lib/express/server.js
```diff
@@ -70,24 +72,44 @@ async function server (configureRoutes, port, hostname = null, allowCors = true)
   });
 }
 
-function configureServer (app, configureRoutes, allowCors = true) {
+function normalizeBasePath (basePath) {
```
KazuCocoa: Is it better to add tests to clarify the cases?

lib/basedriver/driver.js
```diff
@@ -241,11 +249,13 @@ class BaseDriver extends Protocol {
...
+   * lib/protocol, but external packages depend on its existence here. So those
+   * should be updated to rely on the export from lib/protocol before this is
+   * removed.
    */
-  static determineProtocol (desiredCapabilities, requiredCapabilities, capabilities) {
-    return _.isPlainObject(capabilities) ?
-      BaseDriver.DRIVER_PROTOCOL.W3C :
-      BaseDriver.DRIVER_PROTOCOL.MJSONWP;
+  static determineProtocol (...args) {
+    return determineProtocol (...args);
```
mykola-mokhnach: isn't it a recursive call?

lib/basedriver/driver.js
```diff
@@ -520,7 +530,7 @@ class BaseDriver extends Protocol {
       if (!_.isRegExp(avoidPathRegex)) {
         throw new Error('Proxy avoidance path must be a regular expression');
       }
-      let normalizedUrl = url.replace(/^\/wd\/hub/, '');
+      let normalizedUrl = url.replace(new RegExp(`^${this.basePath}`), '');
```
mykola-mokhnach: shouldn't we call _.escapeRegexp?

mykola-mokhnach: same below for other regexps where `this.basePath` is used

lib/protocol/protocol.js
```diff
@@ -7,12 +7,17 @@ import {
...
 import { METHOD_MAP, NO_SESSION_ID_COMMANDS } from './routes';
 import B from 'bluebird';
-import BaseDriver from '../basedriver/driver';
 import {
   MJSONWP_ELEMENT_KEY, W3C_ELEMENT_KEY, formatResponseValue, formatStatus,
 } from './helpers';
 import SESSIONS_CACHE from './sessions-cache';
 
+
+const PROTOCOLS = {
```
mykola-mokhnach: I'd rather set the corresponding values from BaseDriver here. There is no need to duplicate these strings

lib/express/server.js
jlipps: yes, good idea to add some unit tests for this method

lib/basedriver/driver.js
jlipps: it would be if it were `return this.determineProtocol(...args)` but it is calling the imported method, not itself. that's the point--i didn't want to make this a breaking change so i didn't want to remove the method from the BaseDriver class, but i needed the functionality itself to be defined elsewhere. so it is essentially an alias until we can move all the usages of it throughout the appium projects, then we can remove this method entirely.

lib/basedriver/driver.js
jlipps: what is the concern with putting the unescaped string here?

lib/protocol/protocol.js
jlipps: i thought that's what I'm doing. these strings are not being duplicated, they are being moved (though i am keeping the alias in BaseDriver so as not to make this a breaking change for dependents, similar to how i explained for determineProtocol above)

lib/basedriver/driver.js
jlipps: i think the worst that could happen is route parsing stuff would stop working. but there are some valid characters like "." that would not be escaped. so yes i will do this.

lib/basedriver/driver.js
imurchie: Why not make the breaking change and do the work to update the other packages?

PR #350, opened 2019-08-21
refactor: make server parameters a hash [breaking change]

imurchie:
The parameters for starting the server are beginning to get long, so switch to a hash.

If #346 is a made into a breaking change, this can piggy-back onto that, and only need a single major release of the package.

### DIFF
--- a/lib/express/README.md
+++ b/lib/express/README.md
@@ -13,7 +13,8 @@ The `appium-express` server comes configured with:
 4. default error handling
 5. fix for invalid content types sent by certain clients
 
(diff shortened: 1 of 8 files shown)


mykola-mokhnach:
what if we make the change non-breaking by analyzing the type of the first parameter (e.g. if it is a function then use the old style otherwise the new one). This way third party drivers could still receive base driver updates

imurchie:
The change in #346 should legitimately be a breaking change, so this might as well be also, to piggy-back on that and not include a bunch of conditional parameter handling.

mykola-mokhnach:
Burn it

PR #353, opened 2019-08-30
fix: Patch downstream responses for unknown upstream protocol

mykola-mokhnach:
the idea is to add `status` and `sessionId` fields to getStatus responses if the call was made before session creation request has been sent
Addresses https://github.com/appium/appium/issues/13161


lib/jsonwp-proxy/protocol-converter.js
```diff
@@ -199,6 +212,11 @@ class ProtocolConverter {
    */
   async convertAndProxy (commandName, url, method, body) {
     if (!this.downstreamProtocol) {
+      if (commandName === 'getStatus') {
+        // Patch getStatus call for GENERIC protocol
```
KazuCocoa: 👀 
ah...

mykola-mokhnach: There are existing tests, I would say they do cover such changes

PR #357, opened 2019-09-23
fix: Adjust the error codes for NotImplementedError

mykola-mokhnach:
Selenium client is complaining about the error codes we return from these exceptions. See, for example, https://discuss.appium.io/t/exception-in-thread-main-org-openqa-selenium-unsupportedcommandexception-method-has-not-yet-been-implemented/27958

### DIFF
--- a/docs/mjsonwp/errors.md
+++ b/docs/mjsonwp/errors.md
@@ -15,8 +15,8 @@ These classes, which are constructed with a string message (defaulting to the "D
 | 11   | `ElementNotVisibleError`         | An element command could not be completed because the element is not visible on the page
 | 12   | `InvalidElementStateError`       | An element command could not be completed because the element is in an invalid state (e.g., attempting to click a disabled element)
 | 13   | `UnknownError`                   | An unknown server-side error occurred while processing the command
-| 13   | `NotYetImplementedError`         | The operation requested is not yet implemented by the driver
-| 13   | `NotImplementedError`            | The operation requested will not be implemented by the driver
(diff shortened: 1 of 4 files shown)


mykola-mokhnach:
@jlipps Selenium contributors have 405 as the error code: https://github.com/SeleniumHQ/selenium/blob/176b4a9e3082ac1926f2a436eb346760c37a5998/java/client/src/org/openqa/selenium/remote/ErrorCodes.java#L231

jlipps:
@mykola-mokhnach those look like http status codes, not jsonwp status codes. i'm not surprised to not see jsonwp status codes in the selenium codebase, because selenium has upgraded to w3c. my point is that it doesn't make sense to change a status code 13 to 405. the issue is probably that the client is expecting an http status code of 405, but in jsonwp mode we are returning an http status code of 500 with a response object declaring a status field with the value 13.

mykola-mokhnach:
> those look like http status codes

no, this is exactly the status code.If you check the constructor of KnownError class then the first parameter is named jsonCode: https://github.com/SeleniumHQ/selenium/blob/176b4a9e3082ac1926f2a436eb346760c37a5998/java/client/src/org/openqa/selenium/remote/ErrorCodes.java#L287

And the constant used for that error is https://github.com/SeleniumHQ/selenium/blob/176b4a9e3082ac1926f2a436eb346760c37a5998/java/client/src/org/openqa/selenium/remote/ErrorCodes.java#L105

imurchie:
So essentially this is making the two errors be what they said they were, aliases of `UnknownMethodError`, instead of aliases of `UnknownError`.

PR #361, opened 2019-10-12
refactor: Reuse walkDir implementation from appium-support module

mykola-mokhnach:


lib/basedriver/helpers.js
```diff
@@ -257,7 +245,8 @@ async function unzipApp (zipPath, dstRoot, supportedAppExtensions) {
   try {
     logger.debug(`Unzipping '${zipPath}'`);
     await zip.extractAllTo(zipPath, tmpRoot);
-    const allExtractedItems = await walkDir(tmpRoot);
+    const allExtractedItems = [];
+    await fs.walkDir(tmpRoot, true, (itemPath) => void allExtractedItems.push(itemPath));
```
imurchie: One question... how does this differ from using a well-documented-and-tested function like `glob`, which we expose through `appium-support`? It seems to be doing the same thing.

mykola-mokhnach: glob could probably do it as well

PR #364, opened 2019-11-04
feat: add new route to allow logging of custom events

jlipps:
So people can use the event parser timeline in context of their own events!


mykola-mokhnach:
This endpoint should be documented on appium.io

jlipps:
Will add docs!

PR #367, opened 2019-11-07
fix: Fix the request options for different keepAlive values

mykola-mokhnach:
The documentation states that setting `forever` is equal to setting the `agent` to `http(s).Agent({keepAlive:true})`. This means we should not set `agent` to `false` if `forever` is `true`

### DIFF
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -125,7 +125,6 @@ class JWProxy {
       _.isString(content) ? content : JSON.stringify(content),
       { length: LOG_OBJ_LENGTH });
     const reqOpts = {
-      agent: false,
       url: newUrl,
       method,
       headers: {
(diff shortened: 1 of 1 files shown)


lib/jsonwp-proxy/proxy.js
```diff
@@ -135,8 +134,12 @@ class JWProxy {
       },
       resolveWithFullResponse: true,
       timeout: this.timeout,
-      forever: this.keepAlive,
     };
+    if (this.keepAlive) {
+      reqOpts.forever = true;
+    } else {
+      reqOpts.agent = false;
```
imurchie: Are these mutually exclusive? If in the initialization of the `reqOpts` object can it not be
```js
const reqOpts = {
  // ...
  forever: this.keepAlive,
  agent: !this.keepAlive,
}
```

mykola-mokhnach: no it won't work this way, at least for `agent` key. This must be an instance of http.Agent class (`false` is only needed to disable it). Similar situation with `forever`: it implicitly sets the agent value if set to `true`, so we should not explicitly set its value on our own

imurchie: Ok! I think it might do well to be noted, otherwise at some point someone will do the 'optimization' I was thinking of.

mykola-mokhnach: added comments

PR #370, opened 2019-11-25
refactor: Prefer event emitting to promises cancelation

mykola-mokhnach:
According to https://github.com/appium/appium/issues/13631 we still have memory leaks due to weird promise cancelation logic. IMHO promise cancelation itself is evil in its current implementation and I would rather avoid using it. 

BREAKING CHANGE: The current refactoring is breaking and changes `onUnexpectedShutdown` to a normal event, which is emitted by the driver when it is shut down unexpectedly. I've also checked the depending drivers and it looks like only the umbrella driver uses this promise, so we'll have to change that implementation after the major version bump

### DIFF
--- a/lib/basedriver/driver.js
+++ b/lib/basedriver/driver.js
@@ -102,12 +102,12 @@ class BaseDriver extends Protocol {
    * when the driver is shut down unexpectedly. Multiple calls to this method
    * will cause the handler to be executed mutiple times
    *
-   * @param {Function} hanlder The code to be executed on unexpected shutdown.
+   * @param {Function} handler The code to be executed on unexpected shutdown.
    * The function may accept one argument, which is the actual error instance, which
    * caused the driver to shut down.
    */
-  onUnexpectedShutdown (hanlder) {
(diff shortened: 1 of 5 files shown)


lib/basedriver/driver.js
```diff
@@ -91,9 +91,25 @@ class BaseDriver extends Protocol {
...
+  /**
+   * Set a callback handler if needed to execute a custom piece of code
+   * when the driver is shut down unexpectedly. Multiple calls to this method
+   * will cause the handler to be executed mutiple times
+   *
+   * @param {Function} hanlder The code to be executed on unexpected shutdown.
+   * The function may accept one argument, which is the actual error instance, which
+   * caused the driver to shut down.
+   */
+  onUnexpectedShutdown (hanlder) {
```
imurchie: handler

mykola-mokhnach:
@imurchie I can also observe the TODO at https://github.com/appium/appium/blob/67e85ff02ce0d9d1237912e38b207869431e7e07/lib/appium.js#L411

which we could possible resolve by this PR

PR #371, opened 2019-11-26
fix: make sure the listener gets removed after the command executor is completed

mykola-mokhnach:
Should address https://github.com/appium/appium/issues/13631#issuecomment-558505896

### DIFF
--- a/lib/basedriver/driver.js
+++ b/lib/basedriver/driver.js
@@ -323,12 +323,22 @@ class BaseDriver extends Protocol {
       throw new errors.NotYetImplementedError();
     }
 
+    let unexpectedShutdownListener;
     const commandExecutor = async () => imgElId
       ? await ImageElement.execute(this, cmd, imgElId, ...args)
       : await B.race([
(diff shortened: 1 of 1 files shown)


lib/basedriver/driver.js
```diff
@@ -323,12 +323,22 @@ class BaseDriver extends Protocol {
...
       : await B.race([
         this[cmd](...args),
-        new B((resolve, reject) => this.eventEmitter.on(ON_UNEXPECTED_SHUTDOWN_EVENT, reject))
-      ]);
+        new B((resolve, reject) => {
+          unexpectedShutdownListener = reject;
+          this.eventEmitter.on(ON_UNEXPECTED_SHUTDOWN_EVENT, unexpectedShutdownListener);
+        })
+      ]).finally(() => {
+        // https://github.com/appium/appium/issues/13631
```
jlipps: in general, IMO a comment which is only a link is not as useful as one which briefly explains something in words and then also includes a link.

PR #375, opened 2019-12-15
fix: Perform synchronous removal of applications cache

mykola-mokhnach:
Asynchronous calls in onExit handler are not supported., since the event loop does not exist anymore. See https://nodejs.org/api/process.html#process_event_exit

### DIFF
--- a/lib/basedriver/helpers.js
+++ b/lib/basedriver/helpers.js
@@ -81,7 +81,7 @@ function getCachedApplicationPath (link, currentModified) {
     return fullPath;
   }
   logger.debug(`'Last-Modified' timestamp of '${link}' has been updated. ` +
-    `An updated copy of the application is going to be downloaded.`);
+    `The fresh copy of the application is going to be downloaded.`);
   return null;
 }
 
--- a/lib/basedriver/helpers.js
+++ b/lib/basedriver/helpers.js
@@ -44,7 +44,7 @@ process.on('exit', () => {
   APPLICATIONS_CACHE.reset();
 
   logger.debug(`Performing cleanup of ${appPaths.length} cached ` +
-    `application${appPaths.length > 1 ? 's' : ''}`);
+    `application${appPaths.length === 1 ? '' : 's'}`);
   for (const appPath of appPaths) {
     try {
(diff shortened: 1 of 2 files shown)


lib/basedriver/helpers.js
```diff
@@ -34,32 +35,25 @@ const APPLICATIONS_CACHE_GUARD = new AsyncLock();
...
-    })());
-  }
-  try {
-    await B.all(cleanupPromises);
-  } catch (e) {
-    logger.error(e);
-  } finally {
-    APPLICATIONS_CACHE.reset();
+  logger.debug(`Performing cleanup of ${appPaths.length} cached ` +
+    `application${appPaths.length > 1 ? 's' : ''}`);
```
imurchie: This will give the wrong grammar if there are `0` applications.

lib/basedriver/helpers.js
```diff
@@ -34,32 +35,25 @@ const APPLICATIONS_CACHE_GUARD = new AsyncLock();
...
+  for (const appPath of appPaths) {
+    try {
+      // Asynchronous calls are not supported in onExit handler
+      rimraf.sync(appPath);
```
imurchie: I'd rather this get wrapped inside `appium-support#fs`, where the asynchronous `rimraf` is already wrapped.

lib/basedriver/helpers.js
mykola-mokhnach: %), makes sense

lib/basedriver/helpers.js
mykola-mokhnach: What if we merge it as is for now and then I will commit a patch after the corresponding method is added to appium-support?

lib/basedriver/helpers.js
mykola-mokhnach: however in this particular case `appPaths.length` will always be greater than zero because of the preceding guard condition

lib/basedriver/helpers.js
imurchie: Fine with me! Publishing `appium-support` kicks off a lot of CI work, so it's best done on off hours, anyway.

PR #382, opened 2020-01-10
fix: terminate socket on DELETE session and NoSuchDriverError

imurchie:
In the case of ending sessions, and when a request is for a session that does not exist, the connection should not remain alive.

### DIFF
--- a/lib/protocol/protocol.js
+++ b/lib/protocol/protocol.js
@@ -245,10 +245,13 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
     let newSessionId;
     let currentProtocol = extractProtocol(driver, req.params.sessionId);
 
+    let terminateSocket = false;
+
     try {
       // if this is a session command but we don't have a session,
       // error out early (especially before proxying)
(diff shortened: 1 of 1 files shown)


lib/protocol/protocol.js
```diff
@@ -423,8 +427,13 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
       }
 
       httpResBody = formatStatus(httpResBody, httpStatus, currentProtocol);
-
       res.status(httpStatus).json(httpResBody);
+
+      // in certain situations the socket should not be kept alive
+      if (terminateSocket) {
+        SESSIONS_CACHE.getLogger(req.params.sessionId || newSessionId, currentProtocol).debug(`Destroying socket connection`);
+        res.socket.destroy();
```
mykola-mokhnach: should we rather invoke `end` ?

imurchie: `res.end()` simply signals that the response is fully written. Also, the previous call to `res.json()` causes the `end` to get called.

The issue here is that the socket remains, if it is keep-alive.

PR #393, opened 2020-03-15
chore: Add width/height verification to avoid OOM on image resize

mykola-mokhnach:
https://github.com/appium/appium/issues/14060


mykola-mokhnach:
@imurchie The first message tells about `screen size` and the second one about `screenshot size`

imurchie:
Ah. Well, I for one will be incapable of distinguishing these. Alas. :(

PR #395, opened 2020-03-18
fix: Add button to actions payload

jsa34:
Fix for winappdriver click action payload (see https://github.com/appium/appium-windows-driver/pull/69#issuecomment-600772369)

### DIFF
--- a/lib/basedriver/image-element.js
+++ b/lib/basedriver/image-element.js
@@ -176,10 +176,8 @@ class ImageElement {
         parameters: {pointerType: 'touch'},
         actions: [
           {type: 'pointerMove', x, y, duration: 0},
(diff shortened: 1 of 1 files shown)


lib/basedriver/image-element.js
```diff
@@ -176,9 +176,11 @@ class ImageElement {
         parameters: {pointerType: 'touch'},
         actions: [
           {type: 'pointerMove', x, y, duration: 0},
-          {type: 'pointerDown'},
+          //Button added due to https://github.com/appium/appium-windows-driver/pull/69#issuecomment-600772369
```
mykola-mokhnach: the comment is not needed

jsa34: Comments removed

PR #400, opened 2020-03-28
feat: Add requests idempotency verification on the server side

mykola-mokhnach:
The idea behind that is to be able to properly handle client side retries. 

How this is supposed to work:
- The client adds a randomly generated `X-Idempotency-Key` header to each request we'd like to make idempotent (each request must have a different value). The server itself only considers header presence in POST and PUT requests.
- The server temporarily stores the response stream of the marked request in the internal LRU cache. After the request is done then its result is serialised to the file system.
- If the second request with an idempotency key is coming and no matches are found then the request proceeds further
- If there is an active request with the same key then the response stream of that request will be rerouted to the current one, so the client would receive the same response
- If there was a request with the same key then the previously serialized cached response would be piped to the current response
- All cached responses are removed from the cache automatically as soon as they expire or the cache reaches its limit, since its a LRU cache type

Read https://stripe.com/docs/api/idempotent_requests for more details on this topic

### DIFF
--- a/lib/express/idempotency.js
+++ b/lib/express/idempotency.js
@@ -9,6 +9,7 @@ import { EventEmitter } from 'events';
 const CACHE_SIZE = 1024;
 const IDEMPOTENT_RESPONSES = new LRU({
   max: CACHE_SIZE,
+  updateAgeOnGet: true,
   dispose: (key, {response}) => {
     if (response) {
       fs.rimrafSync(response);
--- a/lib/express/idempotency.js
+++ b/lib/express/idempotency.js
@@ -10,7 +10,7 @@ const CACHE_SIZE = 1024;
 const IDEMPOTENT_RESPONSES = new LRU({
   max: CACHE_SIZE,
   updateAgeOnGet: true,
-  dispose: (key, {response}) => {
+  dispose (key, {response}) {
     if (response) {
       fs.rimrafSync(response);
     }
--- a/lib/express/idempotency.js
+++ b/lib/express/idempotency.js
@@ -1,0 +1,133 @@
+import log from './logger';
+import LRU from 'lru-cache';
+import { fs, util } from 'appium-support';
+import os from 'os';
+import path from 'path';
+import { EventEmitter } from 'events';
+
+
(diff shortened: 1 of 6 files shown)


mykola-mokhnach:
This change will require to tune the clients, although I think the improvement worth it

mykola-mokhnach:
cc @ki4070ma

jlipps:
Can you first help me understand in which scenarios we want client side retries?

mykola-mokhnach:
@jlipps We don't control it. Client libraries, like OkHttp (the one that java client uses), apply retries automatically. See https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/-builder/retry-on-connection-failure/ for more details

package.json
```diff
@@ -53,7 +53,7 @@
     "serve-favicon": "^2.4.5",
     "source-map-support": "^0.5.5",
     "validate.js": "^0.13.0",
-    "webdriverio": "^5.10.9",
+    "webdriverio": "^6.0.2",
```
imurchie: Why is this bumped in _this_ PR? Seems unrelated.

lib/express/idempotency.js
```diff
@@ -0,0 +1,134 @@
...
+import os from 'os';
+import path from 'path';
+import { EventEmitter } from 'events';
+
+
+const CACHE_SIZE = 1024;
+const IDEMPOTENT_RESPONSES = new LRU({
+  max: CACHE_SIZE,
+  updateAgeOnGet: true,
+  dispose: (key, {response}) => {
```
imurchie: No need to make an anonymous function here 
```js
const IDEMPOTENT_RESPONSES = new LRU({
  max: CACHE_SIZE,
  updateAgeOnGet: true,
  dispose (key, {response}) {
    if (response) {
      fs.rimrafSync(response);
    }
  },
});
```

package.json
mykola-mokhnach: rebase glitch. Will revert

KazuCocoa:
👍

PR #403, opened 2020-03-31
chore: fix linting issues

imurchie:


lib/basedriver/commands/execute-child.js
```diff
@@ -106,7 +106,7 @@ function coerceScriptResult (obj) {
 
   // in the cae of an array, just recurse into the items
   if (_.isArray(obj)) {
-    return obj.map(i => coerceScriptResult(i));
+    return obj.map((i) => coerceScriptResult(i));
```
mykola-mokhnach: `return obj.map(coerceScriptResult);`

PR #409, opened 2020-05-05
refactor: Use axios for the actual proxy implementation instead of the deprecated request module

mykola-mokhnach:
BREAKING CHANGE:
- The request method of Proxy class now accepts and returns axios-specific values
- The default value of keepAlive setting has been changed to `true`

### DIFF
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -22,7 +22,7 @@ const {MJSONWP, W3C} = PROTOCOLS;
 
 class JWProxy {
   constructor (opts = {}) {
-    Object.assign(this, {
+    _.defaults(this, opts, {
       scheme: 'http',
       server: 'localhost',
       port: 4444,
@@ -31,7 +31,7 @@ class JWProxy {
       sessionId: null,
       timeout: DEFAULT_REQUEST_TIMEOUT,
       keepAlive: true,
-    }, opts);
+    });
     this.scheme = this.scheme.toLowerCase();
     this._activeRequests = [];
     this._reqCancelSource = axios.CancelToken.source();
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -30,7 +30,7 @@ class JWProxy {
       reqBasePath: DEFAULT_BASE_PATH,
       sessionId: null,
       timeout: DEFAULT_REQUEST_TIMEOUT,
-      keepAlive: false,
+      keepAlive: true,
     });
     this.scheme = this.scheme.toLowerCase();
     this._activeRequests = [];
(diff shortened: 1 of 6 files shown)


lib/jsonwp-proxy/proxy.js
```diff
@@ -27,20 +30,32 @@ class JWProxy {
       reqBasePath: DEFAULT_BASE_PATH,
       sessionId: null,
       timeout: DEFAULT_REQUEST_TIMEOUT,
-      keepAlive: false,
+      keepAlive: true,
```
jlipps, line 33: why change this value?

lib/jsonwp-proxy/proxy.js
```diff
@@ -132,91 +145,79 @@ class JWProxy {
...
             : `Got response with unknown status: ${error}`);
         }
       } else {
         log.warn(e.message);
         log.debug(e.stack);
       }
-      throw new errors.ProxyRequestError(`Could not proxy command to remote server. ` +
-        `Original error: ${e.message}`, e.error, e.statusCode);
+      throw new errors.ProxyRequestError(`Could not proxy command to the remote server. ` +
+        `Original error: ${e.message}`, e.response?.data, e.response?.status);
```
jlipps, line 220: oh we can use null coalescing now? yay!

package.json
```diff
@@ -49,8 +49,6 @@
     "lru-cache": "^5.0.0",
```
jlipps, line 49: i don't see axios being included in this file? was it already added?

jlipps:
i'm ok with the breaking change; is it possible for you to summarize the change that would be necessary downstream to accommodate this for users of jwproxy?

package.json
mykola-mokhnach: it is already there (was added in previous PRs)

lib/jsonwp-proxy/proxy.js
mykola-mokhnach: this is available since gulp plugins 5.2.0

lib/jsonwp-proxy/proxy.js
mykola-mokhnach: In general it is better to have keepAlive enabled by default, since it saves time while sending requests to the downstream server. It was previously enabled, but we've disabled it, because of some internal node + system buffers issues. It was long time ago, so I'd expect correct node versions work much better now and should handle buffering properly.

I'll spend some time debugging the server close helper to understand why keepAlive is blocking it

mykola-mokhnach: Actually here is the thing: https://github.com/nodejs/node-v0.x-archive/issues/9066

mykola-mokhnach:
>  is it possible for you to summarize the change that would be necessary downstream to accommodate this for users of jwproxy?

@jlipps Actually, nothing is needed to be changed unless clients use some special stuff returned by the changed private `request` method

PR #411, opened 2020-05-08
refactor: Move common constants to a separate file to avoid unnecessary cross-imports

mykola-mokhnach:


mykola-mokhnach:
No, the change is not breaking. External imports are still the same.

jlipps:
OK, i seem to remember some places where we imported from files inside the base-driver package, but maybe not. and if all exports are available from the main module, it's a good time to update those anyway.

lib/constants.js
```diff
@@ -0,0 +1,27 @@
...
+// behind a /wd/hub prefix. This should be optional and indeed the server
+// administrator should be able to host the API behind any prefix; but for now
+// it is the default.
+const DEFAULT_BASE_PATH = '/wd/hub';
+const IMAGE_ELEMENT_PREFIX = 'appium-image-element-';
+
+
+export {
+  MAX_LOG_BODY_LENGTH, MJSONWP_ELEMENT_KEY, W3C_ELEMENT_KEY,
+  PROTOCOLS, DEFAULT_BASE_PATH, IMAGE_ELEMENT_PREFIX,
```
KazuCocoa, line 26: 👍

PR #418, opened 2020-07-04
fix: extract ipa properly

KazuCocoa:
Tried to fix https://github.com/appium/appium/issues/14100#issuecomment-653733643 without workaround https://github.com/appium/appium/pull/14510

The root cause of the issue is current unzip generates wrong unzipped filenames by Unicode zipped files by Xcode.
When I created a sample app which only added Japanese titled file as a bundled resource, current unzip generated below:

```
$ ls /var/folders/y6/524wp8fx0xj5q1rf6fktjrb00000gn/T/202064-53753-11msanj.01vrk/kazu-sample.app/
Assets.car               Info.plist               _CodeSignature           kazu-sample
Base.lproj               PkgInfo                  embedded.mobileprovision µùÑµ£¼Φ¬₧.png
```

This PR will be:

```
$ ls /var/folders/y6/524wp8fx0xj5q1rf6fktjrb00000gn/T/202064-74785-1yg4iv.4fiis/kazu-sample.app/
Assets.car               Info.plist               _CodeSignature           kazu-sample
Base.lproj               PkgInfo                  embedded.mobileprovision 日本語.png
```

https://github.com/ZJONSSON/node-unzipper#readme

~~I only tested my local ipa, but not for Android and over http yet.~~ <= Tested with our sample apps.

I haven't changed other zip stuff to introduce unzipper from the minimal scope

### DIFF
--- a/lib/basedriver/helpers.js
+++ b/lib/basedriver/helpers.js
@@ -1,3 +1,4 @@
+import os from 'os';
 import _ from 'lodash';
 import path from 'path';
 import url from 'url';
@@ -6,7 +7,7 @@ import { tempDir, fs, util, zip, net } from 'appium-support';
 import LRU from 'lru-cache';
 import AsyncLock from 'async-lock';
 import axios from 'axios';
-import unzipper from 'unzipper';
+import {Open as OpenZip} from 'unzipper';
 
 const ZIP_EXTS = ['.zip', '.ipa'];
 const ZIP_MIME_TYPES = [
@@ -322,10 +323,12 @@ async function unzipApp (zipPath, dstRoot, supportedAppExtensions) {
   try {
     logger.debug(`Unzipping '${zipPath}'`);
     // zip.extractAllTo could not unzip Unicode filenames properly by Xcode
-    // because of https://github.com/appium/appium/issues/14100
-    // so here uses unzipper which can handle the case
-    await unzipper.Open.file(zipPath)
-      .then(d => d.extract({path: tmpRoot, concurrency: 5}));
+    // because of https://github.com/appium/appium/issues/14100.
+    // The unzipper does not keep file permissions (should set as option explicitly),
+    // so we use the library only for unzip the app under test instead of
+    // the replacement of zip.extractAllTo internal.
+    const destDir = await OpenZip.file(zipPath);
+    await destDir.extract({path: tmpRoot, concurrency: os.cpus().length});
 
     const allExtractedItems = await fs.glob('**', {cwd: tmpRoot});
     logger.debug(`Extracted ${util.pluralize('item', allExtractedItems.length, true)} from '${zipPath}'`);
--- a/lib/basedriver/helpers.js
+++ b/lib/basedriver/helpers.js
@@ -6,7 +6,7 @@ import { tempDir, fs, util, zip, net } from 'appium-support';
 import LRU from 'lru-cache';
 import AsyncLock from 'async-lock';
 import axios from 'axios';
-
+import unzipper from 'unzipper';
 
 const ZIP_EXTS = ['.zip', '.ipa'];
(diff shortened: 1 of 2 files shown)


lib/basedriver/helpers.js
```diff
@@ -321,7 +321,12 @@ async function unzipApp (zipPath, dstRoot, supportedAppExtensions) {
   const tmpRoot = await tempDir.openDir();
   try {
     logger.debug(`Unzipping '${zipPath}'`);
-    await zip.extractAllTo(zipPath, tmpRoot);
+    // zip.extractAllTo could not unzip Unicode filenames properly by Xcode
+    // because of https://github.com/appium/appium/issues/14100
+    // so here uses unzipper which can handle the case
+    await unzipper.Open.file(zipPath)
+      .then(d => d.extract({path: tmpRoot, concurrency: 5}));
```
mykola-mokhnach, line 328: we don't need `.then` since Open also returns a promise

lib/basedriver/helpers.js
```diff
@@ -321,7 +321,12 @@ async function unzipApp (zipPath, dstRoot, supportedAppExtensions) {
   const tmpRoot = await tempDir.openDir();
   try {
     logger.debug(`Unzipping '${zipPath}'`);
-    await zip.extractAllTo(zipPath, tmpRoot);
+    // zip.extractAllTo could not unzip Unicode filenames properly by Xcode
```
mykola-mokhnach, line 324: I would rather replace the implementation of `extractAllTo` inside appium-support

lib/basedriver/helpers.js
```diff
@@ -6,7 +6,7 @@ import { tempDir, fs, util, zip, net } from 'appium-support';
 import LRU from 'lru-cache';
 import AsyncLock from 'async-lock';
 import axios from 'axios';
-
+import unzipper from 'unzipper';
```
mykola-mokhnach, line 9: `import { Open as OpenZip } from 'unzipper';`

lib/basedriver/helpers.js
mykola-mokhnach:
```
const directory = await OpenZip.file(...);
await directory.extract(...);
```

mykola-mokhnach: it makes sense to set concurrency equal to the cpu cores count (`os.cpus().length`)

lib/basedriver/helpers.js
KazuCocoa: Yeah, I searched which repositories used `extractAllTo`. Will create a PR to the support after checking

lib/basedriver/helpers.js
KazuCocoa: yey, it is I wanted to find... > os.cpus().length

lib/basedriver/helpers.js
KazuCocoa: I found the unzipper did not keep permission. In chromedriver download feature, unzipper did unzip the binary as 644. It should have been 755. So, I think this unzipper should be used as a spot like this case rather than replacement of general extractAllTo.

mykola-mokhnach: Could you please describe it as a bug to unzipper repository?: https://github.com/ZJONSSON/node-unzipper/issues

Are you sure that extracted files will be working as expected on real/virtual devices after being extracted without original permissions preserved?

KazuCocoa: https://github.com/EvanOxfeld/node-unzip/issues/123 and https://github.com/ZJONSSON/node-unzipper/issues/131 are about the permission/mode stuff.

> Are you sure that extracted files...

Yes, only our test apps and a simple app I created to check the Unicode issue though.

But I think we should keep all files as is, so should close this PR.

PR #422, opened 2020-07-14
feat: allow plugins to modify the server and method map

jlipps:
These changes are needed in order to support the new plugin architecture for Appium 2.0. I do not believe they are breaking changes since they just add optional arguments, and change the shape of a method which I don't think is used outside this repo.

### DIFF
--- a/lib/express/server.js
+++ b/lib/express/server.js
@@ -33,105 +33,42 @@ async function server (opts = {}) {
     hostname = null,
     allowCors = true,
     basePath = DEFAULT_BASE_PATH,
+    plugins = []
   } = opts;
 
   // create the actual http server
   const app = express();
-  let httpServer = http.createServer(app);
-  const serverStateNotifier = new EventEmitter();
-  let isServerClosed = false;
-  httpServer.addWebSocketHandler = addWebSocketHandler;
-  httpServer.removeWebSocketHandler = removeWebSocketHandler;
-  httpServer.removeAllWebSocketHandlers = removeAllWebSocketHandlers;
-  httpServer.getWebSocketHandlers = getWebSocketHandlers;
+  const httpServer = http.createServer(app);
+  return await new B(async (resolve, reject) => {
+    // we put an async function as the promise constructor because we want some things to happen in
+    // serial (application of plugin updates, for example). But we still need to use a promise here
+    // because some elements of server start failure only happen in httpServer listeners. So the
+    // way we resolve it is to use an async function here but to wrap all the inner logic in
+    // try/catch so any errors can be passed to reject.
+    try {
+      configureHttp({httpServer, reject});
+      configureServer({app, addRoutes: routeConfiguringFunction, allowCors, basePath, plugins});
+      await configureServerPlugins({app, httpServer, plugins});
 
-  // http.Server.close() only stops new connections, but we need to wait until
-  // all connections are closed and the `close` event is emitted
-  const close = httpServer.close.bind(httpServer);
-  httpServer.close = async () => await new B((resolve, reject) => {
-    // https://github.com/nodejs/node-v0.x-archive/issues/9066#issuecomment-124210576
-    isServerClosed = true;
-    serverStateNotifier.emit('shutdown');
-    httpServer.on('close', resolve);
-    close((err) => {
-      if (err) reject(err); // eslint-disable-line curly
-    });
-  });
-
-  return await new B((resolve, reject) => {
-    httpServer.on('error', (err) => {
-      if (err.code === 'EADDRNOTAVAIL') {
-        log.error('Could not start REST http interface listener. ' +
-                  'Requested address is not available.');
-      } else {
-        log.error('Could not start REST http interface listener. The requested ' +
-                  'port may already be in use. Please make sure there is no ' +
-                  'other instance of this server running already.');
-      }
-      reject(err);
-    });
-    httpServer.on('connection', (socket) => {
-      socket.setTimeout(KEEP_ALIVE_TIMEOUT_MS);
-      socket.on('error', reject);
-
-      function destroy () {
-        if (socket._openReqCount === 0) {
-          socket.destroy();
-        }
-      }
-      socket._openReqCount = 0;
-      socket.once('close', () => serverStateNotifier.removeListener('shutdown', destroy));
-      serverStateNotifier.once('shutdown', destroy);
-    });
-    httpServer.on('request', function (req, res) {
-      const socket = req.connection || req.socket;
-      socket._openReqCount++;
-      res.on('finish', function () {
-        socket._openReqCount--;
-        if (isServerClosed && socket._openReqCount === 0) {
-          socket.destroy();
-        }
-      });
-    });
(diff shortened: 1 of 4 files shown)


lib/express/server.js
```diff
@@ -93,7 +94,18 @@ async function server (opts = {}) {
...
       });
     });
-    configureServer(app, routeConfiguringFunction, allowCors, basePath);
+
+    configureServer({app, addRoutes: routeConfiguringFunction, allowCors, basePath, plugins});
+
+    // allow plugins to update the app and http server objects
+    for (const plugin of plugins.filter((p) => p.updatesServer)) {
+      log.info(`Allowing plugin ${plugin.name} to modify the Appium server`);
+      plugin.updateServer(app, httpServer).catch(reject);
```
mykola-mokhnach, line 103: should we check first if the plugin has `updateServer` method and this method returns a promise? should this method run asynchronously?

lib/protocol/protocol.js
```diff
@@ -210,13 +210,16 @@ function routeConfiguringFunction (driver) {
     throw new Error('Drivers used with MJSONWP must implement `executeCommand` or `execute`');
   }
 
-  // return a function which will add all the routes to the driver
-  return function addRoutes (app, basePath = DEFAULT_BASE_PATH) {
+  // return a function which will add all the routes to the driver. Here extraMethods might be
+  // passed in as defined by Appium plugins, so we need to add those to the default list
+  return function addRoutes (app, basePath = DEFAULT_BASE_PATH, extraMethods) {
```
mykola-mokhnach, line 215: I would rather make everything after `app` argument as options, so the signature would look like `addRoutes(app, ops = {}) `

lib/express/server.js
```diff
@@ -164,16 +182,13 @@ function configureServer (app, routeConfiguringFunction, allowCors = true, baseP
   // set up start logging (which depends on bodyParser doing its thing)
   app.use(startLogFormatter);
 
-  routeConfiguringFunction(app, basePath);
+  addRoutes(app, basePath, plugins.map((p) => p.newMethodMap));
```
mykola-mokhnach, line 185: should we check first if newMethodMap is a plain object?

test/express/server-specs.js
```diff
@@ -14,8 +14,8 @@ describe('server configuration', function () {
   it('should actually use the middleware', function () {
```
mykola-mokhnach, line 14: would it be possible to add a unit test for the new feature?

lib/express/server.js
jlipps: Yeah, good idea, for better error messages.

lib/protocol/protocol.js
jlipps: 👍

lib/express/server.js
jlipps: 👍

lib/express/server.js
jlipps: Oh, and you're right, we want to run those updates synchronously, so I'll need to refactor this a bit.

jlipps:
ok all changes addressed and unit and e2e tests added

lib/protocol/protocol.js
```diff
@@ -210,13 +210,16 @@ function routeConfiguringFunction (driver) {
     throw new Error('Drivers used with MJSONWP must implement `executeCommand` or `execute`');
   }
 
-  // return a function which will add all the routes to the driver
-  return function addRoutes (app, basePath = DEFAULT_BASE_PATH) {
+  // return a function which will add all the routes to the driver. Here extraMethods might be
+  // passed in as defined by Appium plugins, so we need to add those to the default list
+  return function addRoutes ({app, basePath = DEFAULT_BASE_PATH, extraMethods = {}}) {
```
mykola-mokhnach, line 215: it makes sense to keep `app` outside of the options dictionary, but up to you to decide

test/express/server-specs.js
```diff
@@ -10,15 +10,76 @@ import sinon from 'sinon';
...
+    const driver = fakeDriver();
+    const addRoutes = routeConfiguringFunction(driver);
+    const plugin = fakePlugin();
+    plugin.newMethodMap = [];
+    configureServer({app: app1, addRoutes});
+    configureServer({app: app2, addRoutes, plugins: [plugin]});
+    app2.totalCount().should.eql(app1.totalCount());
+  });
+
+  it('should allow plugins to update the server', async function () {
```
mykola-mokhnach, line 80: should the test be empty?

lib/express/server.js
```diff
@@ -33,13 +33,16 @@ async function server (opts = {}) {
...
   const app = express();
-  let httpServer = http.createServer(app);
-  const serverStateNotifier = new EventEmitter();
-  let isServerClosed = false;
+  const httpServer = http.createServer(app);
+  const serverState = {
+    notifier: new EventEmitter(),
+    closed: false,
+  };
   httpServer.addWebSocketHandler = addWebSocketHandler;
```
mykola-mokhnach, line 46: this all stuff could also be moved to http server configuration

lib/express/server.js
```diff
@@ -50,88 +53,44 @@ async function server (opts = {}) {
...
-                  'other instance of this server running already.');
-      }
+  return await new B(async (resolve, reject) => {
+    // we put an async function as the promise constructor because we want some things to happen in
+    // serial (application of plugin updates, for example). But we still need to use a promise here
+    // because some elements of server start failure only happen in httpServer listeners. So the
+    // way we resolve it is to use an async function here but to wrap all the inner logic in
+    // try/catch so any errors can be passed to reject.
+    try {
+      configureHttp({httpServer, serverState, reject});
```
mykola-mokhnach, line 71: these two synchronous methods could be extracted out of the promise

jlipps: I don't think so, since they need the `reject` method, right?

mykola-mokhnach: ah, I see now. Makes sense

test/express/server-specs.js
jlipps: whoops, forgot to implement that one.

jlipps:
ok, all comments addressed on this PR.

lib/express/server.js
```diff
@@ -164,16 +101,131 @@ function configureServer (app, routeConfiguringFunction, allowCors = true, baseP
...
+async function configureServerPlugins ({plugins, app, httpServer}) {
+  // allow plugins to update the app and http server objects
+  for (const plugin of plugins.filter((p) => p.updatesServer)) {
+    log.info(`Allowing plugin ${plugin.name} to modify the Appium server`);
+    try {
+      // wrap this in error handling in case the plugin forgot to mark the function async or
+      // forgot to implement it
+      await plugin.updateServer(app, httpServer);
+    } catch (err) {
+      log.error(`Plugin '${plugin.name}' tried to update the server but the updateServer ` +
```
mykola-mokhnach, line 191: it makes sense to write the original error stack into debug log here

jlipps: stack will be printed when the original error is thrown in the main appium process

PR #423, opened 2020-08-11
Issue#14637: send empty string to device instead of null when custome…

JesseCodeBones:
…r replace the content with empty value

### DIFF
--- a/lib/protocol/routes.js
+++ b/lib/protocol/routes.js
@@ -581,7 +581,7 @@ const METHOD_MAP = {
       // We want to either a value (old JSONWP) or a text (new W3C) parameter,
       // but only send one of them to the command (not both).
       // Prefer 'value' since it's more backward-compatible.
-      makeArgs: (jsonObj) => [jsonObj.value || jsonObj.text],
+      makeArgs: (jsonObj) => (jsonObj.value || jsonObj.text) == null ? [''] : [jsonObj.value || jsonObj.text],
     }}
   },
   '/session/:sessionId/appium/settings': {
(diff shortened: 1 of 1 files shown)


lib/protocol/routes.js
```diff
@@ -581,7 +581,7 @@ const METHOD_MAP = {
       // We want to either a value (old JSONWP) or a text (new W3C) parameter,
       // but only send one of them to the command (not both).
       // Prefer 'value' since it's more backward-compatible.
-      makeArgs: (jsonObj) => [jsonObj.value || jsonObj.text],
+      makeArgs: (jsonObj) => (jsonObj.value || jsonObj.text) == null ? [''] : [jsonObj.value || jsonObj.text],
```
mykola-mokhnach, line 584: `[jsonObj.value || jsonObj.text || '']` should work

mykola-mokhnach:
@JesseCodeBones  Please sign the CLA

lib/protocol/routes.js
mykola-mokhnach: or even better `[jsonObj.value ?? jsonObj.text]`

JesseCodeBones: Hi Mykola, 
I chosed the first change, if the `value` and `text` are both null value, the null value could also send to device.
so it seems first chose is better.

Regards

mykola-mokhnach: then it would make sense to still replace `||` with `??`

KazuCocoa:
The CLA must be the same as your commit's email address

JesseCodeBones:
rebuildpr

JesseCodeBones:
@KazuCocoa  @mykola-mokhnach  
I resigned the CLA, but the check is not pass, how to retrigger the checks for Pull Request? Thank you.

lib/protocol/routes.js
jessechen0319: @mykola-mokhnach finished the change, please kindly review, Thank you.

PR #424, opened 2020-08-13
chore: Always destroy active sockets on server shutdown

mykola-mokhnach:
This change ensures there will be no unexpected delays when appium process receives SIGINT or SIGTERM


mykola-mokhnach:
yes, the integration test needs to be updated with the new behaviour

PR #425, opened 2020-08-18
fix: allow prefixed standard capabilities

dpgraham:
Currently, if we provide W3C capabilities and prefix standard W3C capabilities (platformName, browserName, etc...) it rejects the session. This PR is made, in the spirit of forgiveness, to allow standard capabilities to be prefixed with `appium:` and let the offenders off with a `log.warn`.

Example of session creation caps that will fail in Appium's current state

```
{
	"capabilities": {
		"firstMatch": [{}],
		"alwaysMatch": {
		  "appium:platformName": "iOS",
		  "appium:browserName": "safari",
		  "appium:automationName": "XCUITest",
		  "appium:deviceName": "iPhone Simulator",
		  "appium:platformVersion": "13.4",
		  "sauce:options": {
		    "appiumVersion": "1.16.0"
		  }
		}
	}
}
```

### DIFF
--- a/lib/basedriver/capabilities.js
+++ b/lib/basedriver/capabilities.js
@@ -87,10 +87,17 @@ function stripAppiumPrefixes (caps) {
     // If it's standard capability that was prefixed, add it to an array of incorrectly prefixed capabilities
     if (isStandardCap(strippedCapName)) {
       badPrefixedCaps.push(strippedCapName);
+      if (_.isNil(caps[strippedCapName])) {
+        caps[strippedCapName] = caps[prefixedCap];
+      } else {
+        log.warn(`Ignoring capability '${prefixedCap}=${caps[prefixedCap]}' and ` +
+          `using capability '${strippedCapName}=${caps[strippedCapName]}'`);
+      }
+    } else {
+      caps[strippedCapName] = caps[prefixedCap];
     }
 
     // Strip out the prefix
-    caps[strippedCapName] = caps[prefixedCap];
     delete caps[prefixedCap];
   }
 
--- a/lib/basedriver/capabilities.js
+++ b/lib/basedriver/capabilities.js
@@ -96,7 +96,7 @@ function stripAppiumPrefixes (caps) {
 
   // If we found standard caps that were incorrectly prefixed, throw an exception (e.g.: don't accept 'appium:platformName', only accept just 'platformName')
   if (badPrefixedCaps.length > 0) {
-    throw new errors.InvalidArgumentError(`The capabilities ${JSON.stringify(badPrefixedCaps)} are standard capabilities and should not have the "appium:" prefix`);
+    log.warn(`The capabilities ${JSON.stringify(badPrefixedCaps)} are standard capabilities and do not require "appium:" prefix`);
   }
 }
 
--- a/lib/basedriver/capabilities.js
+++ b/lib/basedriver/capabilities.js
@@ -198,6 +198,7 @@ function parseCaps (caps, constraints = {}, shouldValidateCaps = true) {
       }
     } catch (err) {
       log.warn(err.message);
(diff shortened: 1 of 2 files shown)


lib/basedriver/capabilities.js
```diff
@@ -96,7 +96,7 @@ function stripAppiumPrefixes (caps) {
 
   // If we found standard caps that were incorrectly prefixed, throw an exception (e.g.: don't accept 'appium:platformName', only accept just 'platformName')
   if (badPrefixedCaps.length > 0) {
-    throw new errors.InvalidArgumentError(`The capabilities ${JSON.stringify(badPrefixedCaps)} are standard capabilities and should not have the "appium:" prefix`);
+    log.warn(`The capabilities ${JSON.stringify(badPrefixedCaps)} are standard capabilities and do not require "appium:" prefix`);
```
KazuCocoa, line 99: here is for standard capabilities.
So I think it should ignore capabilities which have `:` separator.

I mean `alwaysMatch: {'appium:platformName': 'Whatevz'}...` should be rejected, but `alwaysMatch: {'sauce:options': 'Whatevz'}...` should accept. But `alwaysMatch: {'sauce:platformName': 'Whatevz'}...` should be rejected.

mykola-mokhnach:
What would be the final cap value if something like 

```
{
  "appium:platformName": "iOS",
  "platformName": "Android",
}
```

or 

```
		"firstMatch": [{
                  "appium:platformName": "iOS",
                }],
		"alwaysMatch": {
		  "platformName": "iOS"
		}
```

is passed?

dpgraham:
Good thinking @mykola-mokhnach ... I think we ought to defer to unprefixed. I'll make a unit test to cover that.

test/basedriver/capabilities-specs.js
```diff
@@ -248,11 +248,18 @@ describe('caps', function () {
...
-      (() => processCapabilities({
+    it('should still accept prefixed caps even if they are standard capabilities (https://www.w3.org/TR/webdriver/#dfn-table-of-standard-capabilities)', function () {
+      processCapabilities({
         alwaysMatch: {'appium:platformName': 'Whatevz'},
         firstMatch: [{'appium:browserName': 'Anything'}],
-      })).should.throw(/standard capabilities/);
+      }).should.deep.equal({platformName: 'Whatevz', browserName: 'Anything'});
+    });
+
+    it('should prefer standard caps that are non-prefixed to prefixed', function () {
```
mykola-mokhnach, line 258: it would also be good to have a test for 

```
"firstMatch": [{
  "appium:platformName": "iOS"
}],
"alwaysMatch": {
  "platformName": "Android"
}
```

case

KazuCocoa: And with other prefixes as your original motivation?
```
"sauce:options": {
  "appiumVersion": "1.16.0"
}
```

dpgraham: @mykola-mokhnach I just confirmed that having duplicate keys in firstMatch and alwaysMatch is forbidden.

From the documentation:
```
For each enumerable own property in secondary, run the following substeps:

Let name be the name of the property.

Let value be the result of getting a property named name from secondary.

Let primary value be the result of getting the property name from primary.

If primary value is not undefined, **return an error with error code invalid argument**.

Set a property on result with name name and value value.
```

(primary value refers to the value that came from `alwaysMatch` object)

dpgraham: @KazuCocoa other non appium: prefixes just get ignored. The prefixes aren't stripped out.

KazuCocoa (approved):

dpgraham:
@mykola-mokhnach I added a test case to confirm that it errors with a proper message when alwaysMatch and firstMatch contain duplicates.

PR #429, opened 2020-08-27
feat: Remove wd hub prefix

KazuCocoa:
https://github.com/appium/appium/projects/2#card-43906187
We want to remove `/wd/hub` default base path.

--

Haven't tested and tweaked well yet, but did the task intend to do like this? @jlipps 
(This PR is to 2.0 branch)

### DIFF
--- a/docs/mjsonwp/protocol-methods.md
+++ b/docs/mjsonwp/protocol-methods.md
@@ -8,78 +8,78 @@ See the WebDriver [W3C](https://w3c.github.io/webdriver/webdriver-spec.html#list
 
 HTTP Method | Path                                                                   | Details
 ------------|------------------------------------------------------------------------|---------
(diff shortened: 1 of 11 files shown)


jlipps:
@KazuCocoa i think we should merge this into the 2.0 branch of base driver and publish as `@beta` or `@next`, sound good?

jlipps:
> we need to update the default for the -ap arg on the main server too.

actually we don't need to do this as long as we update the exported DEFAULT_BASE_PATH in this PR

KazuCocoa:
Yeah, could be 🆗

jlipps:
@KazuCocoa can you remove the WIP tag so we can merge?

KazuCocoa:
removed ✅

jlipps:
ok merged and published as beta!

PR #432, opened 2020-09-08
fix: Return proper responses to unexpected errors

mykola-mokhnach:
All clients expect responses to be returned as valid json objects and throw exceptions in case of plain text: https://gist.github.com/TransmitTomerDavid/31132cd2ea3ce3c4514e24494ad8f504

### DIFF
--- a/lib/express/middleware.js
+++ b/lib/express/middleware.js
@@ -50,7 +50,11 @@ function defaultToJSONContentType (req, res, next) {
   next();
 }
 
-function catchAllHandler (err, req, res) {
+function catchAllHandler (err, req, res, next) {
+  if (res.headersSent) {
+    return next(err);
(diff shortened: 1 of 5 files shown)


jlipps (approved):
LGTM, but pretty sure this could break JSONWP client support?

mykola-mokhnach:
@jlipps According to what I see in clients all of them are only expecting JSON responses.

mykola-mokhnach:
@jlipps Could you also help fixing https://travis-ci.org/github/appium/appium-base-driver/jobs/725344616#L1804 (perhaps in the other PR). I have a feeling these tests were broken before.

jlipps:
sure, will take a look

KazuCocoa (approved):

PR #434, opened 2020-09-28
fix: catchall handler should be added after plugin application

jlipps:
somewhere along the way plugins were broken. Maybe a change in Express? Anyway, need to put the catchall handler at the very end, after any plugin might be registered.

### DIFF
--- a/lib/express/server.js
+++ b/lib/express/server.js
@@ -50,6 +50,11 @@ async function server (opts = {}) {
       configureServer({app, addRoutes: routeConfiguringFunction, allowCors, basePath, plugins});
       await configureServerPlugins({app, httpServer, plugins});
 
+      // once all configurations and plugins have been applied, make sure to set up a catchall
+      // handler so that anything unknown 404s. But do this after everything else since we don't
+      // want to block plugins' ability to add routes if they want.
+      app.all('*', catch404Handler);
+
       await startServer({httpServer, hostname, port});
       resolve(httpServer);
     } catch (err) {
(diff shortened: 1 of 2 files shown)


lib/express/server.js
```diff
@@ -50,6 +50,11 @@ async function server (opts = {}) {
       configureServer({app, addRoutes: routeConfiguringFunction, allowCors, basePath, plugins});
       await configureServerPlugins({app, httpServer, plugins});
 
+      // once all configurations and plugins have been applied, make sure to set up a catchall
+      // handler so that anything unknown 404s. But do this after everything else since we don't
+      // want to block plugins' ability to add routes if they want.
+      app.all('*', catch404Handler);
```
mykola-mokhnach, line 56: I'm not quite sure why this change is needed. The actual plugin routes are added in line 105 (`configureServer` function)

jlipps: it's because there are two ways for plugins to update routes. one is via the extra routes way, and one is 'manually' so to speak, by messing with the express app object. for example here: https://github.com/appium/appium-fake-plugin/blob/master/lib/plugin.js#L25-L28

mykola-mokhnach: ok, makes sense. IMHO, I would only keep routes update as the only possible way to do that, because the other way is too implicit.

mykola-mokhnach (approved):

lib/express/server.js
jlipps: i understand that perspective. unfortunately since we are giving them the app object, they can do whatever they want with it, and it seems better to support that than have some cases fail silently in weird ways.

jlipps:
published in 7.3.1

PR #438, opened 2020-11-27
chore: merge master to 2.0 branch

KazuCocoa:
to up to date

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "keywords": [
     "appium"
   ],
-  "version": "7.3.1",
+  "version": "7.3.3",
   "author": "appium",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "keywords": [
     "appium"
   ],
-  "version": "7.3.1",
+  "version": "7.3.3",
```
jlipps, line 7: should we make this 8.0.0-beta? or something? on this branch?

jlipps (approved):

package.json
KazuCocoa: good idea. This PR only has 'merge master' diff, so it should be another PR like such bumping version :)

PR #443, opened 2020-12-07
fix: 10 minute connection timeout is too low for some use cases

jlipps:
Someone reported a bug where they tried to background the app for 30 minutes and it didn't work. Now we know why!

### DIFF
--- a/lib/express/server.js
+++ b/lib/express/server.js
@@ -23,7 +23,7 @@ import { DEFAULT_BASE_PATH } from '../constants';
 import { EventEmitter } from 'events';
 
 
-const KEEP_ALIVE_TIMEOUT_MS = 60 * 10 * 1000; // 10 minutes
+const KEEP_ALIVE_TIMEOUT_MS = 10 * 60 * 1000; // 10 minutes
 
 
 async function server (opts = {}) {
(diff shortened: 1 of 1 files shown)


lib/express/server.js
```diff
@@ -23,7 +23,7 @@ import { DEFAULT_BASE_PATH } from '../constants';
 import { EventEmitter } from 'events';
 
 
-const KEEP_ALIVE_TIMEOUT_MS = 60 * 10 * 1000; // 10 minutes
+const KEEP_ALIVE_TIMEOUT_MS = 60 * 60 * 1000; // 60 minutes
```
KazuCocoa, line 26: Is it better to leave a comment we should keep this timeout (not to make this short in future change)?

jlipps: yeah probably. we might also eventually want to make it configurable via a cli arg if a long timeout is bad for some people

mykola-mokhnach (changes requested):
I would not do that. The idea is that this timeout should not be greater than the default OS timeout, which is 7200 seconds by default in most of the popular operating systems https://www.veritas.com/support/en_US/article.100028680

Consider making it configurable rather than changing the default value.

jlipps:
@mykola-mokhnach it is still less than 7200 secs. but making it configurable is a good idea for sure.

jlipps:
ok @mykola-mokhnach i made it configurable. now will update the main appium server with an arg

mykola-mokhnach (approved):

jlipps:
published in 7.4.0

PR #444, opened 2020-12-07
fix: 10 minute connection timeout is too low for some use cases

jlipps:
2.0 version of https://github.com/appium/appium-base-driver/pull/443


jlipps:
just going to keep merging master in for now.

PR #445, opened 2020-12-17
[wip] feat: remove mjsonwp from responses

KazuCocoa:
Part of _remove support for non-w3c sessions_ https://github.com/appium/appium/projects/2#card-41526139

- Do not add `MJSONWP_ELEMENT_KEY` as responses
- Do not add `status` as responses

So far, this PR keeps `duplicateKeys` in a case the backend is MJSONWP like chromedriver.


mykola-mokhnach:
I would recommend updating all e2e tests in all appium modules first that use wd and JWP protocol to use W3C instead.

KazuCocoa:
yeah... I took time a bit today, but this was so tough than I imagined...

KazuCocoa:
close for now. Need do other places first

PR #451, opened 2021-01-19
feat: expose base driver version via inherited drivers

jlipps:
This way we can always determine basedriver compatibility between appium server and drivers

### DIFF
--- a/lib/basedriver/driver.js
+++ b/lib/basedriver/driver.js
@@ -3,6 +3,7 @@ import {
 } from '../protocol';
 import { PROTOCOLS, DEFAULT_BASE_PATH } from '../constants';
 import os from 'os';
+import path from 'path';
 import commands from './commands';
 import * as helpers from './helpers';
 import log from './logger';
@@ -20,6 +21,7 @@ B.config({
   cancellation: true,
 });
 
+const BASEDRIVER_VER = require(path.resolve(__dirname, '..', '..', '..', 'package.json')).version;
 const NEW_COMMAND_TIMEOUT_MS = 60 * 1000;
 
 const EVENT_SESSION_INIT = 'newSessionRequested';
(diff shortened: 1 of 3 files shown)


KazuCocoa (approved):

lib/basedriver/driver.js
```diff
@@ -30,6 +32,12 @@ const ON_UNEXPECTED_SHUTDOWN_EVENT = 'onUnexpectedShutdown';
 
 class BaseDriver extends Protocol {
 
+  /**
+   * Make the basedriver version available so for any driver which inherits from this package, we
+   * know which version of basedriver it inherited from
+   */
+  static baseVersion = BASEDRIVER_VER;
```
mykola-mokhnach, line 39: is `static` supported in ECMA standard that Appium uses?

lib/basedriver/driver.js
```diff
@@ -20,6 +21,7 @@ B.config({
   cancellation: true,
 });
 
+const BASEDRIVER_VER = require(path.resolve(__dirname, '..', '..', '..', 'package.json')).version;
```
mykola-mokhnach, line 24: chromedrivers uses a bit different approach: https://github.com/appium/appium-chromedriver/blob/78b0ed6ae4d973783733eb626ae14b5b2edc716c/lib/utils.js#L9

which one is right?

lib/basedriver/driver.js
jlipps: it was already in the list of our babel transpilation plugins. i merely had to add it to our linter as well.

lib/basedriver/driver.js
jlipps: right now i'm assuming that everything is run transpiled. if we're ever in a scenario where that's not the case, the test will tell us.

jlipps: I think chromedriver cares because the install script needs to run untranspiled

jlipps:
published in 8.0.0-beta.3

PR #453, opened 2021-02-02
fix!: remove image element proxy check

jlipps:
and allow proxyRouteIsAvoided to inspect the body (supports https://github.com/appium/appium/pull/15078)


lib/protocol/protocol.js
```diff
@@ -10,25 +10,13 @@ import B from 'bluebird';
 import {
   formatResponseValue, formatStatus,
 } from './helpers';
-import {
-  MJSONWP_ELEMENT_KEY, W3C_ELEMENT_KEY, MAX_LOG_BODY_LENGTH, PROTOCOLS,
-  DEFAULT_BASE_PATH, IMAGE_ELEMENT_PREFIX,
-} from '../constants';
+import { MAX_LOG_BODY_LENGTH, PROTOCOLS, DEFAULT_BASE_PATH } from '../constants';
```
mykola-mokhnach, line 13: do we still need the unused constants?

lib/basedriver/driver.js
```diff
@@ -470,10 +470,11 @@ class BaseDriver extends Protocol {
    * should be made available to overridden methods.
    * @param {string} method - HTTP method of the route
    * @param {string} url - url of the route
+   * @param {object} body - webdriver request body
```
mykola-mokhnach, line 473: I assume the type of body should be something like `{?*}` or it is always an object?

jlipps: yeah good point

lib/basedriver/driver.js
```diff
@@ -470,10 +470,11 @@ class BaseDriver extends Protocol {
    * should be made available to overridden methods.
    * @param {string} method - HTTP method of the route
    * @param {string} url - url of the route
+   * @param {*?} body - webdriver request body
```
mykola-mokhnach, line 473: `{*?}` -> `{?*}`

mykola-mokhnach (approved):

lib/basedriver/driver.js
jlipps: oh, i thought optionality was designated after the type, probably thinking of some other language like typescript maybe.

jlipps:
published as 2.0.0-beta.5

PR #459, opened 2021-02-23
refactor!: generalize concepts of server updater and extra method map

jlipps:
basedriver shouldn't really know about plugins or drivers.
it should just expose the ability to update the server or add methods.

so let's get it out of here and move knowledge about drivers and plugins back to the main appium server where it belongs!


mykola-mokhnach (approved):

KazuCocoa (approved):

jlipps:
published in 8.0.0-beta.6

PR #460, opened 2021-02-24
fix: timeout of JWProxy for httpAgent and httpsAgent

KazuCocoa:
Fix https://github.com/appium/appium/issues/15132

This was when we modify keepalive stuff in https://github.com/appium/appium-base-driver/pull/416/files .
Should the timeout respect the constructor's one?

### DIFF
--- a/lib/jsonwp-proxy/proxy.js
+++ b/lib/jsonwp-proxy/proxy.js
@@ -41,7 +41,6 @@ class JWProxy {
       keepAlive: opts.keepAlive ?? true,
       maxSockets: 10,
       maxFreeSockets: 5,
-      timeout: this.timeout,
     };
     this.httpAgent = new http.Agent(agentOpts);
     this.httpsAgent = new https.Agent(agentOpts);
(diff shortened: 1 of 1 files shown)


lib/jsonwp-proxy/proxy.js
```diff
@@ -41,7 +41,7 @@ class JWProxy {
       keepAlive: opts.keepAlive ?? true,
       maxSockets: 10,
       maxFreeSockets: 5,
-      timeout: 60000,
+      timeout: this.timeout,
```
mykola-mokhnach, line 44: I assume this timeout has no effect, because it gets overridden in reqOpts. I would say it is safe to remove it completely to avoid confusion

KazuCocoa: Okay, thanks. I read the documentation, but haven't decided which was (keep or remove it) better

mykola-mokhnach (approved):

mykola-mokhnach:
Reported the issue to axios: https://github.com/axios/axios/issues/3659

PR #471, opened 2021-03-30
feat: add option to skip app uninstall irrespective of reset type

swaraliRaut:
Adding a new capability "skipAppUninstall" to address https://github.com/appium/appium/issues/14607

### DIFF
--- a/lib/basedriver/commands/session.js
+++ b/lib/basedriver/commands/session.js
@@ -67,7 +67,7 @@ commands.createSession = async function createSession (jsonwpDesiredCapabilities
     this.opts.noReset = false;
   }
   this.opts.fastReset = !this.opts.fullReset && !this.opts.noReset;
-  this.opts.skipUninstall = this.opts.fastReset || this.opts.noReset;
+  this.opts.skipUninstall = this.opts.skipAppUninstall || this.opts.fastReset || this.opts.noReset;
 
   // Prevents empty string caps so we don't need to test it everywhere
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
I would rather not add this capability to the base driver as it might not be applicable for inherited drivers. Consider adding it to particular platform drivers instead

swaraliRaut:
@mykola-mokhnach  "skipUninstall" is already initialized in base driver so I thought utilising that might be better than adding a new opt in platform specific drivers. Adding it to platform specific drivers will lead to duplicate changes in multiple drivers too. eg android driver, ios driver etc. Am I missing something here?

mykola-mokhnach:
I am ok with having duplicated caps. It is better than having caps with no use

swaraliRaut:
opts.skipUninstall already exists in base driver and is being used in majority of derived drivers already. All this does is use the cap to drive the value of already present opts.skipUninstall in base driver.

I would really like to understand your perspective here to help me learn for future contributions, so I would appreciate if you can explain the downside of adding the cap to base driver in this case and why is it preferable to add these changes to multiple derived drivers instead.?

mykola-mokhnach:
Everything that is added to the base driver is them automatically exposed to all other depending drivers. Not all of them (actually only some) have the concept of app to uninstall. Also, it is much easier to change/remove that concept from the particular driver rather from the base driver (such change would require bumping of its major version and updating of all depending modules).

swaraliRaut:
oh, I see. thanks for explaining that to me. I will create separate PRs for particular drivers.

PR #486, opened 2021-05-18
chore: add support for driver cli args

ayvnkhan:


lib/basedriver/driver.js
```diff
@@ -48,6 +48,9 @@ class BaseDriver extends Protocol {
     this.originalCaps = null; // To give the original capabilities to reset
     this.helpers = helpers;
 
+    //allow for cli args to be used by drivers
```
jlipps, line 51: nit: `// ` (with space after)

jlipps (approved):
this looks good. note that because of the monorepo work that @boneskull is doing, anything going into the 2.0 branch on this repo may need to go into appium/appium instead

lib/basedriver/driver.js
```diff
@@ -38,7 +38,7 @@ class BaseDriver extends Protocol {
    */
   static baseVersion = BASEDRIVER_VER;
 
-  constructor (opts = {}, shouldValidateCaps = true) {
+  constructor (opts = {}, driverArgs = {}, shouldValidateCaps = true) {
```
mykola-mokhnach, line 41: this might break existing drivers. Consider adding a new argument to the end rather than in the middle

mykola-mokhnach (changes requested):
a potential breaking change is present

mykola-mokhnach (approved):

PR #487, opened 2021-05-20
chore: remove support for -custom locator strategy(should be handled …

ayvnkhan:
…by plugins instead)


ayvnkhan:
Sorry committed a bit too early, theres some more dependancies I need to remove. Will commit again

lib/basedriver/commands/find.js
```diff
@@ -1,174 +0,0 @@
...
-/**
- * Find an element using a custom plugin specified by the customFindModules cap.
- *
- * @param {string} selector - the selector which the plugin will use to find
- * elements
- * @param {boolean} multiple - whether we want one element or multiple
- *
- * @returns {WebElement} - WebDriver element or list of elements
- */
-commands.findByCustom = async function findByCustom (selector, multiple) {
```
jlipps, line 72: we don't want to delete the whole file. only this function and ones immediately related to it. all the other find element stuff we want to keep in here.

lib/basedriver/driver.js
```diff
@@ -378,18 +378,6 @@ class BaseDriver extends Protocol {
     }
   }
 
-  validateLocatorStrategy (strategy, webContext = false) {
```
jlipps, line 381: i don't think we want to remove this

jlipps (changes requested):

PR #489, opened 2021-05-21
fix: convert element.visualization to string

KazuCocoa:
Calling `toString('base64')` should only when the visualization is not string and has value.
https://github.com/appium/appium-plugins/runs/2636842428?check_suite_focus=true

### DIFF
--- a/lib/basedriver/commands/images.js
+++ b/lib/basedriver/commands/images.js
@@ -81,7 +81,7 @@ commands.compareImages = async function compareImages (mode, firstImage, secondI
  *
  **/
 function convertVisualizationToBase64 (element) {
-  if (_.isString(element.visualization)) {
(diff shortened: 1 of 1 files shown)


lib/basedriver/commands/images.js
```diff
@@ -81,7 +81,7 @@ commands.compareImages = async function compareImages (mode, firstImage, secondI
  *
  **/
 function convertVisualizationToBase64 (element) {
-  if (_.isString(element.visualization)) {
+  if (!_.isEmpty(element.visualization) && !_.isString(element.visualization)) {
```
mykola-mokhnach, line 84: I assume !_isEmpty condition should be enough

mykola-mokhnach (approved):

PR #495, opened 2021-06-03
feat: add driver args to base driver

ayvnkhan:


mykola-mokhnach:
I was thinking about whether we can attach this `driverArgs` property dynamically to the driver instance inside of the Appium umbrella driver. Then there will be no need to change the signature of the constructor.

ayvnkhan:
> I was thinking about whether we can attach this `driverArgs` property dynamically to the driver instance inside of the Appium umbrella driver. Then there will be no need to change the signature of the constructor.

PR #498, opened 2021-06-11
feat: allow instrumentation via base driver

Dharin-shah:
Instrumentation Object Schema:
 ```
{
"eventName" : "initializeDriver",
"duration" : 12220
"options" : {},
"children" : [{
                        "eventName" : "driverServerStart",
                        "duration": 8760,
                        "options" : {},
                        "children": []
                     }]
}
```

This data will help when building custom spans using opentelemetry plugin as well as works as a meta field for detailed instrumentation (which can be kept optional)

### DIFF
--- a/lib/basedriver/driver.js
+++ b/lib/basedriver/driver.js
@@ -179,6 +179,38 @@ class BaseDriver extends Protocol {
     return _.cloneDeep(this._meta._instrumentation.data);
   }
 
+  wrapInstrumentation (eventName, fn) {
+    const spannerFunction = (...args) => {
+      this.instrumentEventStart(eventName);
+      let result;
+      try {
+        result = fn.call(...args);
+        this.instrumentEventEnd(eventName);
+      } catch (error) {
+        this.instrumentEventEnd(eventName);
+        throw error;
+      }
+      return result;
+    };
+    return spannerFunction;
+  }
+
+  wrapInstrumentationAsync (eventName, fn) {
+    const spannerFunction = async (...args) => {
+      this.instrumentEventStart(eventName);
+      let result;
+      try {
+        result = await fn.call(...args);
+        this.instrumentEventEnd(eventName);
+      } catch (error) {
+        this.instrumentEventEnd(eventName);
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by jlipps — COMMENTED** [review]
the code itself looks good but i'm not sure what the overall purpose of the _instrumentation object is so I'm finding it hard to do a complete review

**Comment by KazuCocoa** [comment]
Test code like `_eventHistory` for instrumentation also may help to understand the usage.

**Comment by Dharin-shah** [comment]
@KazuCocoa Yes will add the required tests. Also @jlipps updated description for more info. Will add the tests for usage clarity

File: lib/basedriver/driver.js [hunk 1]
```diff
@@ -88,9 +87,16 @@ class BaseDriver extends Protocol {
     // allow subclasses to have internal drivers
     this.managedDrivers = [];
 
-    // store event timings
-    this._eventHistory = {
-      commands: [] // commands get a special place
+    this._meta = {
+      _eventHistory: {
```
jlipps, line 91: given that `_meta` is already "private", its subfields don't really need to be (i.e., we can get rid of `_` here I think)

lib/basedriver/driver.js
```diff
@@ -160,15 +166,96 @@ class BaseDriver extends Protocol {
...
+      this.instrumentEventStart(eventName);
+      let result;
+      try {
+        result = fn.call(...args);
+        this.instrumentEventEnd(eventName);
+      } catch (error) {
+        this.instrumentEventEnd(eventName);
+        throw error;
+      }
+      return result;
```
jlipps, line 193: if we return in the try, we don't need the separate `let` instantiation.

Dharin-shah: done

lib/basedriver/driver.js
Dharin-shah: done

mykola-mokhnach:
Closing this PR as we are not going to maintain this module anymore. It has been moved to `@appium` monorepo

PR #503, opened 2021-07-13
feat(helpers): leverage system unzip where available

boneskull:
In my testing, I've found that `/usr/bin/unzip` is a 3x speedup over a native-JS implementation (which we currently use).  This can be significant in large files.

To maintain cross-platform compatibility, the `unzipApp()` helper function will fall back to the native JS implementation where `/usr/bin/unzip` is not available (Windows).  It will also retry using the native implementation if `/usr/bin/unzip` fails for any other reason.  Given `.zip` files are _validated_ prior to this call, it's unlikely that would occur unless there is something wrong with system unzip itself.

I've added a couple trivial tests here, but to test it "properly" would require a fair bit of refactoring.

TODO: Port to `@appium/base-driver` and configure CI to run E2E tests on win32.

### DIFF
--- a/lib/basedriver/helpers.js
+++ b/lib/basedriver/helpers.js
@@ -2,11 +2,10 @@ import _ from 'lodash';
 import path from 'path';
 import url from 'url';
 import logger from './logger';
-import { tempDir, fs, util, zip, net, timing } from 'appium-support';
+import { system, tempDir, fs, util, zip, net, timing } from 'appium-support';
 import LRU from 'lru-cache';
 import AsyncLock from 'async-lock';
 import axios from 'axios';
-
 
 const IPA_EXT = '.ipa';
 const ZIP_EXTS = ['.zip', IPA_EXT];
@@ -323,7 +322,15 @@ async function unzipApp (zipPath, dstRoot, supportedAppExtensions) {
   try {
     logger.debug(`Unzipping '${zipPath}'`);
     const timer = new timing.Timer().start();
-    const extractionOpts = {};
+    /**
+     * Attempt to use use the system `unzip` (e.g., `/usr/bin/unzip`) due
+     * to the significant performance improvement it provides over the native
+     * JS "unzip" implementation.
+     * @type {import('appium-support/lib/zip').ExtractAllOptions}
+     */
+    const extractionOpts = {
+      useSystemUnzip: !system.isWindows(),
+    };
     // https://github.com/appium/appium/issues/14100
     if (path.extname(zipPath) === IPA_EXT) {
       logger.debug(`Enforcing UTF-8 encoding on the extracted file names for '${path.basename(zipPath)}'`);
--- a/test/basedriver/helpers-specs.js
+++ b/test/basedriver/helpers-specs.js
@@ -1,7 +1,8 @@
-import { isPackageOrBundle, duplicateKeys, parseCapsArray } from '../../lib/basedriver/helpers';
+import { zip, fs, tempDir } from 'appium-support';
+import { configureApp, isPackageOrBundle, duplicateKeys, parseCapsArray } from '../../lib/basedriver/helpers';
 import chai from 'chai';
 import chaiAsPromised from 'chai-as-promised';
-
+import sinon from 'sinon';
 
 chai.use(chaiAsPromised);
 const should = chai.should();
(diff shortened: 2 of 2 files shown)


boneskull:
The builds need to be updated to `sudo apt install unzip`.

For example, a 555MB file will take ~68s to extract on my MBP using native JS and ~21s using `/usr/bin/unzip`.

mykola-mokhnach (changes requested):
It was working like that before, but was changed to use the node-based zip library. The reason for it is the difference in client environment and thus the difference between platform-dependent zip implementations (e.g. how they handle Unicode filenames). Also, we have apps caching, which only makes unzip necessary for the first time until the app is cached.

boneskull:
See appium/appium-support#226

I will update this PR to leverage that change.

However...

- Use of system unzip is now opt-in.
- A user needs some way to opt-in...
- ...which leads me to believe this needs a new capability, because I don't see anywhere else it could go

boneskull:
...and unfortunately, the `configureApp()` helper does not currently accept any options, so I'm going to have to add another parameter.

boneskull:
driver options (not caps) are a thing... I'm not sure where to specify them, though.

boneskull:
maybe that's the settings API? even then, it's too late, since we need the flag before the session is created

KazuCocoa:
Capabilities or an argument as appium cli (https://github.com/appium/appium/blob/1e30207ec4e413c64396420fbb0388392e88cc54/lib/parser.js) are ways to set something before creating a new session. Others are after establishing a session for now.

jlipps:
In this case, I would say it should be a capability. `usePlatformUnzip` for example. All capabilities are merged onto `this.opts` in the driver.

mykola-mokhnach:
Don't make it too complicated. Simply enable system zip usage here by default for apps extraction. In the worst case it would fallback to the node-based implementation.

boneskull:
@mykola-mokhnach Oh, sorry, I misread/misunderstood what you said.  Yes, it will default to system and fallback to native.  I will make these changes.

boneskull:
I have no idea why the build is breaking... seems unrelated but will confirm.

boneskull:
nope, was my fault--needed a sandbox for sinon, though I'm not entirely sure why...

lib/basedriver/helpers.js
```diff
@@ -323,7 +322,12 @@ async function unzipApp (zipPath, dstRoot, supportedAppExtensions) {
   try {
     logger.debug(`Unzipping '${zipPath}'`);
     const timer = new timing.Timer().start();
-    const extractionOpts = {};
+    /**
+     * @type {ExtractAllOptions}
```
mykola-mokhnach, line 326: is this type notation needed?

mykola-mokhnach: I'd rather leave a comment here about why we default to system unzip

boneskull: yes, it's needed if you care about intellisense, but it's also wrong.

boneskull:
@mykola-mokhnach there was an outstanding change request here, but it seems to be out-of-date.

mykola-mokhnach:
Please bump the minimum appium-support version in package.json

lib/basedriver/helpers.js
```diff
@@ -323,7 +322,15 @@ async function unzipApp (zipPath, dstRoot, supportedAppExtensions) {
...
     const timer = new timing.Timer().start();
-    const extractionOpts = {};
+    /**
+     * Attempt to use use the system `unzip` (e.g., `/usr/bin/unzip`) due
+     * to the significant performance improvement it provides over the native
+     * JS "unzip" implementation.
+     * @type {import('appium-support/lib/zip').ExtractAllOptions}
+     */
+    const extractionOpts = {
+      useSystemUnzip: !system.isWindows(),
```
mykola-mokhnach, line 332: On Windows we could also use PowerShell for such purpose: https://blog.netwrix.com/2018/11/06/using-powershell-to-create-zip-archives-and-unzip-files/

I could create a PR later

test/basedriver/helpers-specs.js
```diff
@@ -1,7 +1,8 @@
-import { isPackageOrBundle, duplicateKeys, parseCapsArray } from '../../lib/basedriver/helpers';
+import { zip, fs, tempDir } from 'appium-support';
+import { configureApp, isPackageOrBundle, duplicateKeys, parseCapsArray } from '../../lib/basedriver/helpers';
```
mykola-mokhnach, line 2: this line seems to be too long

mykola-mokhnach (approved):

test/basedriver/helpers-specs.js
boneskull: imo we shouldn't concern ourselves with formatting changes (unless it fails lint) if the formatting is not automated (e.g., precommit hook); there are bigger fish to fry

(the formatting _is_ automated in the `appium/appium` monorepo)

lib/basedriver/helpers.js
boneskull: this is a good idea, but might not be worth adding unless people want it?

PR #541, opened 2021-11-18
fix: Update apps cache behaviour for expired items

mykola-mokhnach:
After longer meditation on LRU cache sources it turns out that `has` check does not actually care about expired items removal. Only the `get` check does that.
Also, it could be that the internal structure of the cached file/folder is changed while it's being cached. For now I've added a very simple (and fast) integrity check for cached folders, which makes sure that the amount of items in the folder (files and subfolders) has not been decreased since the last check.

Related to https://github.com/appium/appium/issues/16109

### DIFF
--- a/lib/basedriver/helpers.js
+++ b/lib/basedriver/helpers.js
@@ -267,14 +267,15 @@ async function configureApp (app, supportedAppExtensions) {
       archiveHash = await calculateFileIntegrity(archivePath);
       if (archiveHash === cachedAppInfo?.archiveHash) {
         const {fullPath} = cachedAppInfo;
-        if (await fs.exists(fullPath)) {
+        if (await isAppIntegrityOk(fullPath, cachedAppInfo?.integrity)) {
           if (archivePath !== app) {
             await fs.rimraf(archivePath);
           }
           logger.info(`Will reuse previously cached application at '${fullPath}'`);
           return verifyAppExtension(fullPath, supportedAppExtensions);
         }
-        logger.info(`The application at '${fullPath}' does not exist anymore. Deleting it from the cache`);
+        logger.info(`The application at '${fullPath}' does not exist anymore ` +
+          `or its integrity has been damaged. Deleting it from the cache`);
         APPLICATIONS_CACHE.del(app);
       }
       const tmpRoot = await tempDir.openDir();
--- a/lib/basedriver/helpers.js
+++ b/lib/basedriver/helpers.js
@@ -18,13 +18,14 @@ const CACHED_APPS_MAX_AGE = 1000 * 60 * 60 * 24; // ms
 const APPLICATIONS_CACHE = new LRU({
   maxAge: CACHED_APPS_MAX_AGE, // expire after 24 hours
   updateAgeOnGet: true,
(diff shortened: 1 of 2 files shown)


lib/basedriver/helpers.js
```diff
@@ -18,13 +18,14 @@ const CACHED_APPS_MAX_AGE = 1000 * 60 * 60 * 24; // ms
...
-      return;
-    }
-
-    logger.info(`The application '${app}' cached at '${fullPath}' has expired`);
-    await fs.rimraf(fullPath);
+  dispose: (app, {fullPath}) => {
+    logger.info(`The application '${app}' cached at '${fullPath}' has ` +
+      `expired after ${CACHED_APPS_MAX_AGE}ms`);
+    setTimeout(0, async () => {
+      if (fullPath && await fs.exists(fullPath)) {
```
jlipps, line 25: any exception thrown in here will become an unhandled rejection

jlipps: and why are we using setTimeout anyway?

jlipps (approved):

lib/basedriver/helpers.js
mykola-mokhnach: that's ok, both function calls are safe

mykola-mokhnach: setTimeout is used in order to have async calls

jlipps: why not just

```js
(async() => { /*function body*/ })();
```?

Comment (mykola-mokhnach) [reply lib/basedriver/helpers.js:25]: dunno, I find setTimeout more obvious

Comment (jlipps) [reply lib/basedriver/helpers.js:25]: hmm. i guess for me there's nothing that prevents you from running an async function just like you would a regular function if you're not going to wait on the result. but it's fine how it is.

File: lib/basedriver/helpers.js [hunk 2]
```diff
@@ -121,6 +124,28 @@ function verifyAppExtension (app, supportedAppExtensions) {
...
+async function calculateFileIntegrity (filePath) {
+  return await fs.hash(filePath);
+}
+
+async function isAppIntegrityOk (currentPath, expectedIntegrity = {}) {
+  if (!await fs.exists(currentPath)) {
+    return false;
+  }
+  const isDir = (await fs.stat(currentPath)).isDirectory();
+  if (isDir && await calculateFolderIntegrity(currentPath) >= expectedIntegrity?.folder) {
```
KazuCocoa, line 140: so, does this mean current folder size can be greater than the cached one..?
I thought this case also basically compare cached one and current one with `===`

mykola-mokhnach: Yes. this is for the case OS devices to add its own hidden caches there, for example TimeMachine or Finder ones

KazuCocoa: ah, I see.
Maybe a comment about this will help us when we take a look around this code again in the future

KazuCocoa (approved):
👍

lib/basedriver/helpers.js
mykola-mokhnach: added a comment there

PR #542, opened 2021-11-22
fix: Rearrange arguments for setTimeout call

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/16146


KazuCocoa (commented):
👍

